PR #6, opened 2022-01-03
Fix/restructure

Svastits:


src/kuka_sunrise/configuration_manager.cpp
```diff
@@ -51,13 +73,16 @@ ConfigurationManager::ConfigurationManager(
...
-    robot_manager_node_->declare_parameter("controller_ip", rclcpp::ParameterValue("192.168.38.8"));
-  }
-  robot_manager_node_->set_on_parameters_set_callback(
+
+  param_callback_ = robot_manager_node_->add_on_set_parameters_callback(
     [this](
       const std::vector<rclcpp::Parameter> & parameters)
     {return this->onParamChange(parameters);});
+
+  if (!robot_manager_node_->has_parameter("controller_ip")) {
```
kovacsge11, line 82: Why did you put the ip param setting after callback setting?

src/kuka_sunrise/configuration_manager.cpp
```diff
@@ -30,6 +30,28 @@ ConfigurationManager::ConfigurationManager(
...
+      false, false});
+  parameter_set_access_rights_.emplace(
+    "send_period_ms", ParameterSetAccessRights {false, true,
+      false, false, false});
+  parameter_set_access_rights_.emplace(
+    "receive_multiplier", ParameterSetAccessRights {false, true,
+      false, false, false});
+  parameter_set_access_rights_.emplace(
+    "controller_ip", ParameterSetAccessRights {false, false,
+      false, false, true});
```
kovacsge11, line 53: Does the program stand in configuring state long enough for you to be able to set the ip?

src/kuka_sunrise/configuration_manager.cpp
kovacsge11: Is it so the onParamChange will be called with the declaration?

kovacsge11: also, havent you missed setting the actual value?

src/kuka_sunrise/configuration_manager.cpp
kovacsge11: Answer: yes, from cmd line with ros2 node run cmd or from launch file.

src/kuka_sunrise/configuration_manager.cpp
kovacsge11: Answer: we need this to be able to change the value from the already set param value.

kovacsge11 (approved):

PR #13, opened 2022-04-08
Suggestion: mention distributed, real-time control systems explicitly

gavanderhoorn:
At least I believe that's the intent of the text, based on my understanding of the goals of this project.

This is of course just a suggestion from a random stranger.

Feel free to ignore.


resizoltan:
Dear @gavanderhoorn, your feedback is always much appreciated. Support of distributed systems is of course an important aspect, thank you for pointing this out.

PR #16, opened 2022-04-22
Organize packages and add documentation

resizoltan:


resizoltan:
For some reason pull request analysis is not working properly with sonar, but it is an issue with our industrial_ci and not this repo, so I am merging this and I will open an issue there.

PR #19, opened 2022-05-06
Feature/joint controller interface

### LINKED ISSUE #17: Use base class of kroshu_ros2_core
When I took a look at the sonar issues, I realized that we have the ROS2BaseClass class in [this](https://github.com/kroshu/kroshu_ros2_core) repo, which we don't use here at the moment. We could benefit from it, e.g. it would reduce code duplication and introduce greater modularity. I'll take a look at it and give it a go on a separate branch.

### LINKED ISSUE #18: Create base class for joint_controller
The node sending the commands to the sunrise driver must have a few members and functions, which are necessary for proper functioning:
- Control loop with send_period_ms_ and receive_multiplier_ member variables
- Server for syncing these variables with the driver
- lifecycle management and parameter handling (should come from ROS2BaseNode)

A base class with some pure virtual functions should be implemented, which contains these must-have functionalities

Svastits:
Created an interface (JointControllerBase class) for joint_controller with all necessary members and functions to communicate with the sunrise driver.
The class inherits from ROS2BaseNode.
Closes #18 and partly resolves #17

Child classes:
 - InterpolatingContoller: this has the same functionality, as the first JointController (subscribing to reference joint states and interpolating to the velocity limits)
 - RateScaledController (child of InterpolatingController): given a source with const rate for reference joint states, the interpolation is done so, that the motion of the robot finishes when the next reference arrives (if the max velocity allows this)


kovacsge11 (changes requested):
Everything OK, minor remark on service interface.

kovacsge11 (approved):

resizoltan:
Can you please delete the `ParameterSetAccessRights` struct definition from `joint_controller_base.hpp`? It is already defined in the `ROS2BaseNode`.

kuka_sunrise_control/robot_control/src/robot_control/rate_scaled_controller.cpp
```diff
@@ -0,0 +1,184 @@
...
+ScaledJointController::ScaledJointController(
+  const std::string & node_name,
+  const rclcpp::NodeOptions & options)
+: InterpolatingController(node_name, options)
+{
+  auto set_rate_callback = [this](
+    kuka_sunrise_interfaces::srv::SetDouble::Request::SharedPtr request,
+    kuka_sunrise_interfaces::srv::SetDouble::Response::SharedPtr response) {
+      int cmd_per_frame = static_cast<int>(JointControllerBase::ms_in_sec_ /
+        loopPeriod() / (8 * request->data)) + 1;
```
resizoltan, line 34: What is the magic number 8? Publish frequency of camera? If yes, this hidden dependency should be generalized out somehow.

kuka_sunrise_control/robot_control/include/robot_control/rate_scaled_controller.hpp
```diff
@@ -0,0 +1,62 @@
...
+#include "kuka_sunrise_interfaces/srv/set_double.hpp"
+#include "kuka_sunrise_interfaces/srv/set_int.hpp"
+
+#include "robot_control/interpolating_controller.hpp"
+
+
+namespace robot_control
+{
+
+class ScaledJointController : public InterpolatingController
```
resizoltan, line 35: What about
```suggestion
class RateScaledJointController : public InterpolatingController
```
?

kovacsge11:
@resizoltan please re-review.

resizoltan (approved):

PR #21, opened 2022-05-23
Fix/refactor control logic

Svastits:
Changed to inherit from ROS2BaseLCNode to reduce code duplications and fixed some sonar code smells


kovacsge11 (changes requested):
We might need to add exec dependency. But if not, then ok.

kovacsge11 (approved):

PR #22, opened 2022-05-25
Feature/rollback

Svastits:
Rollback is simplified for configuration and activation of the system_manager
- if one component fails, the cleanup/deactivation callback is called for the system manager
- these are implemented so, that they check the actual state of the components and do the rollback if necessary
- this way the nodes should be always in sync (if the rollback fails, error state is triggered)

Changed sync client requests to new SendRequest<> wrapper

### DIFF
--- a/kuka_sunrise_control/teleop_guided_robot/package.xml
+++ b/kuka_sunrise_control/teleop_guided_robot/package.xml
@@ -3,7 +3,7 @@
 <package format="3">
   <name>teleop_guided_robot</name>
   <version>0.0.0</version>
-  <description>Package for managing the connection between ROS and KUKA Sunrise FRI</description>
+  <description>Package for starting keyboard-controlled motion and synchronising the necessary nodes</description>
   <maintainer email="resizoltan@gmail.com">rosdeveloper</maintainer>
   <license>Apache 2.0</license>
 
--- a/kuka_sunrise_control/teleop_guided_robot/package.xml
+++ b/kuka_sunrise_control/teleop_guided_robot/package.xml
@@ -17,6 +17,7 @@
   <depend>std_srvs</depend>
   <depend>kuka_sunrise_interfaces</depend>
   <depend>kuka_sunrise</depend>
+  <depend>kroshu_ros2_core</depend>
   
   <exec_depend>robot_control</exec_depend>
   <exec_depend>pluginlib</exec_depend>
--- a/kuka_sunrise_control/teleop_guided_robot/package.xml
+++ b/kuka_sunrise_control/teleop_guided_robot/package.xml
@@ -3,30 +3,23 @@
 <package format="3">
   <name>teleop_guided_robot</name>
   <version>0.0.0</version>
-  <description>TODO: Package description</description>
+  <description>Package for managing the connection between ROS and KUKA Sunrise FRI</description>
   <maintainer email="resizoltan@gmail.com">rosdeveloper</maintainer>
-  <license>TODO: License declaration</license>
+  <license>Apache 2.0</license>
 
   <buildtool_depend>ament_cmake</buildtool_depend>
 
-  <build_depend>rclcpp</build_depend>
-  <build_depend>rclcpp_lifecycle</build_depend>
-  <build_depend>sensor_msgs</build_depend>
-  <build_depend>geometry_msgs</build_depend>
-  <build_depend>std_msgs</build_depend>
-  <build_depend>std_srvs</build_depend>
-  <build_depend>kuka_sunrise</build_depend>
+  <depend>rclcpp</depend>
+  <depend>rclcpp_lifecycle</depend>
+  <depend>sensor_msgs</depend>
+  <depend>geometry_msgs</depend>
+  <depend>std_msgs</depend>
+  <depend>std_srvs</depend>
+  <depend>kuka_sunrise_interfaces</depend>
+  <depend>kuka_sunrise</depend>
   
-  <exec_depend>rclcpp</exec_depend>
-  <exec_depend>rclcpp_lifecycle</exec_depend>
-  <exec_depend>sensor_msgs</exec_depend>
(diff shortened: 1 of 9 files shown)


kuka_sunrise_driver/src/kuka_sunrise/configuration_manager.cpp
```diff
@@ -404,23 +404,17 @@ bool ConfigurationManager::setCommandMode(const std::string & control_mode)
     RCLCPP_ERROR(robot_manager_node_->get_logger(), "Invalid control mode");
     return false;
   }
-  auto future_result = command_mode_client_->async_send_request(request);
```
kovacsge11, line 407: I don't know where to put it, but we should refactor parameter handling in this source file, too. Right?

kuka_sunrise_driver/src/kuka_sunrise/robot_manager_node.cpp
```diff
@@ -97,19 +97,14 @@ RobotManagerNode::on_configure(const rclcpp_lifecycle::State &)
...
-    std::chrono::milliseconds(3000));
-  if (future_status != std::future_status::ready) {
-    RCLCPP_ERROR(get_logger(), "Future status not ready, could not set parameters");
-    return FAILURE;
-  }
-  if (!future_result.get()->success) {
-    RCLCPP_ERROR(get_logger(), "Future result not success, could not set parameters");
+
+  auto response = kuka_sunrise::sendRequest<std_srvs::srv::Trigger::Response>(
+    set_parameter_client_, trigger_request, 0, 1000);
```
kovacsge11, line 102: Why did you change timeout ms from 3000 to 1000? It occurs multiple times in this PR. (I mean it's ok, but I don't know whether you changed it on purpose.)

kuka_sunrise_control/teleop_guided_robot/package.xml
```diff
@@ -3,9 +3,9 @@
 <package format="3">
   <name>teleop_guided_robot</name>
   <version>0.0.0</version>
-  <description>TODO: Package description</description>
+  <description>Package for managing the connection between ROS and KUKA Sunrise FRI</description>
```
kovacsge11, line 6: But this is not the purpose of the teleop_guided_robot package, is it?

kovacsge11 (changes requested):
Review not finished yet, system_manager class needs review.

kuka_sunrise_driver/src/kuka_sunrise/robot_manager_node.cpp
Svastits: For the changes of simple values (member variables or FRI parameters), 1000ms is more than enough, so I changed them on purpose.
Only the requests of state changes need a longer time, so I left those on the original 3000 ms

kuka_sunrise_control/teleop_guided_robot/package.xml
Svastits: True, I will update this
But I think we should consider moving the system_manager to another package, as it has a more general purpose than just starting the teleop_guided motion

kuka_sunrise_driver/src/kuka_sunrise/configuration_manager.cpp
Svastits: Yes, that should be done, but I would do it in a separate PR, as the current ParameterSetAccessRights structure of the base node is not sufficient for this purpose:
- in this case, the ConfigurationManager is instantiated while the node is in configuring (transition) state, but currently only primary states are supported in that struct
- either the struct should be modified to enable transition states or the code restructured -> FSM restructuring should be planned before

kuka_sunrise_control/teleop_guided_robot/src/control_system/system_manager.cpp
```diff
@@ -14,107 +14,211 @@
...
+  {
     return ERROR;
   }
-  if (!changeState(JOINT_CONTROLLER, lifecycle_msgs::msg::Transition::TRANSITION_DEACTIVATE)) {
+  if (getState(JOINT_CONTROLLER).id != lifecycle_msgs::msg::State::PRIMARY_STATE_INACTIVE &&
+    !changeState(
+      JOINT_CONTROLLER,
+      lifecycle_msgs::msg::Transition::TRANSITION_DEACTIVATE))
+  {
+    return ERROR;
```
kovacsge11, line 204: why use ERROR here and FAILURE at other places?

kuka_sunrise_control/teleop_guided_robot/src/control_system/system_manager.cpp
```diff
@@ -14,107 +14,211 @@
...
 {
-  (void)state;
-  if (!changeState(ROBOT_INTERFACE, lifecycle_msgs::msg::Transition::TRANSITION_CLEANUP)) {
-    return FAILURE;
+  if (getState(ROBOT_INTERFACE).id != lifecycle_msgs::msg::State::PRIMARY_STATE_UNCONFIGURED &&
+    !changeState(
+      ROBOT_INTERFACE,
+      lifecycle_msgs::msg::Transition::TRANSITION_CLEANUP))
+  {
+    return ERROR;
```
kovacsge11, line 126: same here, why ERROR instead of FAILURE?

kovacsge11 (approved):

kovacsge11 (changes requested):
ERROR/FAILURE not clear

kuka_sunrise_control/teleop_guided_robot/src/control_system/system_manager.cpp
Svastits: For deactivation and cleanup, I felt it is safer to return error if transition is not successful, in these cases, the node will go to error state and get finalized or return to unconfigured.

For configure or activate, this is not the desired behaviour, it is safe to return to the original state and thus enable retrying the transition

kovacsge11 (approved):

PR #24, opened 2022-06-10
Fix/param handling

Svastits:
I have changed nodes to inherit from _ROS2BaseLCNode_ for better parameter handling and to reduce code duplication.
Additionally, I modified the rollback of the _robot_manager_node_, so that if the parameter setting in the configuration fails, the nodes transition back to unconfigured.
Changes to the parameter handling:
- added check on IP validity
- the _on_configure_ callback first declares the _controller_ip_ parameter, than connects to the given IP
- if successfully connected, the _robot_manager_  sends a trigger request to declare the other parameters (some of which need connection to the FRI)
    - this way the parameter callback can be registered at first, and the FRI related parameters are declared and set after successful connection -> this is in accordance with the intended ROS2 parameter design

### DIFF
--- a/kuka_sunrise_control/teleop_guided_robot/include/control_system/system_manager.hpp
+++ b/kuka_sunrise_control/teleop_guided_robot/include/control_system/system_manager.hpp
@@ -55,6 +55,7 @@ public:
 
 private:
   bool changeState(const std::string & node_name, std::uint8_t transition);
+  bool changeRobotCommandingState(bool is_active);
   void robotCommandingStateChanged(bool is_active);
   void getFRIState();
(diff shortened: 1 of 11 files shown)


Svastits:
Additional comment: to further improve rollback, the on_error callback should be used, as currently errors are not handled and from error state the node goes to unconfigured state WITHOUT executing the on_cleanup method.
This means that if the node is active and an error occurs during deactivation, the node will transition to unconfigured state, although parts of the driver may remain active
-> the internal state machine for every transition should be seriously considered @resizoltan

resizoltan (approved):

PR #25, opened 2022-06-10
Restructure of the RSI driver

oprezz:
Restructure of the current RSI driver implementation

Rationale:
- Make the HW interface ROS independent
- Create a ROS compatible node as a wrapper for the HW interface

Acceptance:
- [x] Test with real HW
- [x] Test parameter configurability


kuka_rsi_hw_interface/src/robot_control_node.cpp
```diff
@@ -22,76 +22,134 @@
...
-on_configure(
-  const rclcpp_lifecycle::State &)
-{
-  kuka_rsi_hw_interface_ = std::make_unique<KukaHardwareInterface>(this->shared_from_this());
-  kuka_rsi_hw_interface_->configure();
-  return SUCCESS;
-}
+  CallbackReturn RobotControlNode::on_configure(const rclcpp_lifecycle::State &)
+  {
+    registerParameter<std::string>("rsi_ip_address_", "127.0.0.1",
```
Svastits, line 46: ip and port should not be allowed to be changed in inactivate state (2. value in ParameterSetAccessRights should be false)

kuka_rsi_hw_interface/src/robot_control_node.cpp
```diff
@@ -22,76 +22,134 @@
...
-  while (kuka_rsi_hw_interface_->isActive()) {
-    if (!kuka_rsi_hw_interface_->read()) {
-      RCLCPP_ERROR(get_logger(), "Failed to read state from robot. Shutting down!");
-      rclcpp::shutdown();
-      return;
+  void RobotControlNode::ControlLoop()
+  {
+    while (kuka_rsi_hw_interface_->isActive())
+    {
+      std::unique_lock<std::mutex> lock(m_);
```
Svastits, line 97: I would not put the mutex lock before the read() function, as it is a blocking call and thus commandReceivedCallback may not be allowed to run at all

kuka_rsi_hw_interface/src/robot_control_node.cpp
Svastits: also true for n_nof

kuka_rsi_hw_interface/src/kuka_hardware_interface.cpp
```diff
@@ -17,141 +17,101 @@
...
-    return;
-  }
-  for (size_t i = 0; i < n_dof_; i++) {
-    joint_pos_correction_deg_[i] = (joint_command_msg_->position[i] - initial_joint_pos_[i]) *
-      KukaHardwareInterface::R2D;
+  void KukaHardwareInterface::start(std::vector<double> &joint_state_msg_position,
+                                    std::vector<double> &joint_command_msg_position,
+                                    std::vector<double> &initial_joint_position)
+  {
+    std::unique_lock<std::mutex> lock(m_);
```
Svastits, line 37: consider lock_guard instead (theoretically less overhead)

kuka_rsi_hw_interface/include/kuka_rsi_hw_interface/kuka_hardware_interface.hpp
```diff
@@ -28,57 +28,42 @@
...
+               std::vector<double> &joint_command_msg_position,
+               std::vector<double> &initial_joint_position);
+    void stop(const std::vector<double> &joint_pos_correction_deg_);
+    bool read(std::vector<double> &joint_state_msg_position);
+    bool write(const std::vector<double> &joint_pos_correction_deg_);
+
+    bool isActive() const;
+
+  private:
+    unsigned int n_dof_ = 6;
```
Svastits, line 48: maybe declare as const and initialize in constructor

kuka_rsi_hw_interface/include/kuka_rsi_hw_interface/kuka_hardware_interface.hpp
```diff
@@ -28,57 +28,42 @@
...
+    bool is_active_ = false;
+
+    std::vector<std::string> joint_names_ = std::vector<std::string>(6);
```
Svastits, line 51: use n_dof_ instead of 6

kuka_rsi_hw_interface/include/kuka_rsi_hw_interface/robot_control_node.hpp
```diff
@@ -23,32 +23,52 @@
...
-  rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn
-  on_activate(const rclcpp_lifecycle::State &) override;
-  rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn
-  on_deactivate(const rclcpp_lifecycle::State &) override;
+    std::string rsi_ip_address_ = "";
+    int rsi_port_ = 0;
+    unsigned int n_dof_ = 6;
 
-  std::unique_ptr<KukaHardwareInterface> kuka_rsi_hw_interface_;
+    std::vector<double> initial_joint_pos_ = std::vector<double>(6, 0.0);
```
Svastits, line 57: use n_dof_, resize in configure if necessary (by parameter changes)

kuka_rsi_hw_interface/src/robot_control_node.cpp
```diff
@@ -22,66 +22,129 @@
...
-}  // namespace kuka_rsi_hw_interface
 
+void RobotControlNode::commandReceivedCallback(sensor_msgs::msg::JointState::SharedPtr msg)
+{
+  std::lock_guard<std::mutex> lock(m_);
+  joint_command_msg_ = msg;
+  cv_.notify_one();
+}
+
+bool RobotControlNode::onRSIIPAddressChange(const std::string & rsi_ip_address)
```
Svastits, line 129: maybe check IP validity?

Svastits: I would declare all parameters in constructor, as for now, they can't be changed in unconfigured state

kuka_rsi_hw_interface/src/robot_control_node.cpp
```diff
@@ -22,66 +22,129 @@
...
   return SUCCESS;
 }
-
-rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn RobotControlNode::
-on_activate(
-  const rclcpp_lifecycle::State &)
+CallbackReturn RobotControlNode::on_activate(const rclcpp_lifecycle::State &)
 {
-  kuka_rsi_hw_interface_->start();
+  kuka_rsi_hw_interface_->start(
```
Svastits, line 72: I would remove the joint_command argument and set it in the node after the start function

kuka_rsi_hw_interface/include/kuka_rsi_hw_interface/robot_control_node.hpp
```diff
@@ -17,38 +17,57 @@
...
+  void ControlLoop();
 
   std::thread control_thread_;
-};
 
+  std::string rsi_ip_address_ = "";
+  int rsi_port_ = 0;
+  unsigned int n_dof_ = 6;
+
+  std::vector<double> initial_joint_pos_ = std::vector<double>(6, 0.0);
```
Svastits, line 57: consider moving this to the HWinterface, calculate difference and degree conversion there

kuka_rsi_hw_interface/include/kuka_rsi_hw_interface/kuka_hardware_interface.hpp
oprezz: Actually this is not used, and probably not even needed. Deleted.

PR #45, opened 2023-01-18
Feature/rox driver joint imperdance refactoring

altex111:
**Joint impedance controller refactoring:**
- Joint number is configurable from yaml file
- Stiffness and damping parameters are now in diferent order (stiffnes; damping) (stiffness; damping) ...

**Robot manager node refactoring:**
- Control mode parameter and used controller names configurable via yaml file


kuka_rox_hardware_interface/include/kuka_rox_hw_interface/robot_manager_node.hpp
```diff
@@ -64,6 +66,15 @@ class RobotManagerNode : public kroshu_ros2_core::ROS2BaseLCNode
   std::shared_ptr<rclcpp_lifecycle::LifecyclePublisher<std_msgs::msg::Bool>> is_configured_pub_;
   std_msgs::msg::Bool is_configured_msg_;
 
+  const std::string POSITION_CONTROL = "position";
+  const std::string TORQUE_CONTROL = "torque";
+  const std::string IMPEDANCE_CONTROL = "impedance";
+
+  const std::string POSITION_CONTROLLER_NAME = "position_controller_name";
+  const std::string IMPEDANCE_CONTROLLER_NAME = "impedance_controller_name";
+  const std::string TORQUE_CONTROLLER_NAME = "torque_controller_name";
```
VX792, line 75: shouldn't this be called TORQUE_CONTROLLER_NAME_PARAM? As I see, it isn't the name itself, but the name of the parameter which defines the controller's name.

kuka_controllers/include/kuka_controllers/joint_impedance_controller.hpp
```diff
@@ -56,6 +58,22 @@ class JointImpedanceController : public controller_interface::ControllerInterfac
...
+  // Storing command joint names for interfaces
+  std::vector<std::string> command_joint_names_;
+
+  // Parameters from ROS for joint_impedance_controller
+  std::shared_ptr<joint_impedance_controller::ParamListener> param_listener_;
+  joint_impedance_controller::Params params_;
+
+  // Private consts
+  const std::vector<std::string> command_interfaces_param = {"stiffness", "damping"};
+  static constexpr double STIFFNESS_DEFAULT = 10;
```
VX792, line 74: in the sampler and the ecs, the default stiffness is 30, but it probably doesn't matter here

VX792 (changes requested):

kuka_controllers/include/kuka_controllers/joint_impedance_controller.hpp
altex111: @Svastits  whats your opinion?

kuka_controllers/src/joint_impedance_controller.cpp
```diff
@@ -83,11 +135,11 @@ controller_interface::return_type JointImpedanceController::update(
...
 {
 
-  for (auto index = 0; index < 6; ++index) {
+  for (auto index = 0; index < dof_*command_interfaces_param.size(); index+2) {
     command_interfaces_[index].set_value(stiffness_[index]);
   }
-  for (auto index = 6; index < 12; ++index) {
-    command_interfaces_[index].set_value(damping_[index-6]);
+  for (auto index = 1; index < dof_*command_interfaces_param.size(); index+2) {
+    command_interfaces_[index].set_value(damping_[index-dof_]);
```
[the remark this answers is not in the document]
altex111: Corrected it

kuka_rox_hardware_interface/include/kuka_rox_hw_interface/robot_manager_node.hpp
altex111: good suggestion, thanks

kuka_controllers/src/joint_impedance_controller.cpp
```diff
@@ -55,14 +69,52 @@ const
...
+  // Check if the DoF has changed
+  if ((dof_ > 0) && (params_.joints.size() != dof_))
+  {
+    RCLCPP_ERROR(
+      get_node()->get_logger(),
+      "The JointImpedanceController does not support restarting with a different number of DOF");
+    return CallbackReturn::FAILURE;
+  }
+
+  if (params_.joints.empty())
```
Svastits, line 94: is it ok if configuration is successful in this case? We should maybe discuss this

kuka_rox_hardware_interface/src/robot_manager_node.cpp
```diff
@@ -207,6 +241,23 @@ RobotManagerNode::on_deactivate(const rclcpp_lifecycle::State &)
   }
   return SUCCESS;
 }
+
+bool RobotManagerNode::onControlModeChangeRequest(const std::string & control_mode)
+{
+  if(control_mode == POSITION_CONTROL ||
+     control_mode == IMPEDANCE_CONTROL ||
+     control_mode == TORQUE_CONTROL)
```
Svastits, line 249: Do not allow torque, will be only supported later

kuka_controllers/src/joint_impedance_controller.cpp
```diff
@@ -83,11 +135,11 @@ controller_interface::return_type JointImpedanceController::update(
   const rclcpp::Duration &)
 {
 
-  for (auto index = 0; index < 6; ++index) {
```
Svastits, line 86: if you change the order of the interfaces here, you should also change it in the hardware interface, otherwise they will be mixed up

kuka_rox_hardware_interface/src/robot_manager_node.cpp
```diff
@@ -144,15 +171,22 @@ RobotManagerNode::on_activate(const rclcpp_lifecycle::State &)
...
-  // auto impedance_controller_name = this->get_parameter("impedance_controller_name").as_string();
-  // controller_name_ = (this->get_parameter("command_mode").as_string() == "position")
-  //     ? position_controller_name : torque_controller_name;
-  controller_names_.clear();
-  controller_names_.emplace_back("joint_trajectory_controller");
-  if (is_joint_imp_contr_) {
-    controller_names_.emplace_back("joint_impedance_controller");
+  // Select control modes
+  auto position_controller_name = this->get_parameter(POSITION_CONTROLLER_NAME_PARAM).as_string();
+  auto impedance_controller_name = this->get_parameter(IMPEDANCE_CONTROLLER_NAME_PARAM).as_string();
```
Svastits, line 176: With this map structure I think it would be a little better to fill the map when the name changes occur, that way you do not have to fill all fields of the map unnecessarily.

I know we agreed on this way previously, but if we assume that no one changes the controller names just for fun, filling the map only in the name changes is a little bit more efficient in my opinion.

In the future we may also want to allow control mode changes if external control is active, then this solution has too much overhead.
What are your thoughts @altex111 @VX792 @kovacsge11 ?

Svastits (changes requested):

kuka_rox_hardware_interface/src/robot_manager_node.cpp
VX792: This function is only called once so performance-wise it might not be a problem, but I think it would make more sense to do this in the on_configure() function. For one,  reading parameters and creating a data structure which stores a part of the predefined yaml has more to do with configuration than activation, and we also aren't delaying the first message by doing stuff like dynamically allocating vectors and filling the map.

But I think you're right, the solution depends on how changing the control mode in active state will be implemented

kuka_rox_hardware_interface/src/robot_manager_node.cpp
VX792: Do we actually need this function? This can easily be inlined into the lambda where it's called.

kuka_controllers/src/joint_impedance_controller.cpp
VX792: Is this order better though? I personally prefer the other one, since you can tell at a glance if a parameter is damping or stiffness based on which half of the vector they are in. Probably not important enough to address right now though.

VX792 (changes requested):

kuka_controllers/src/joint_impedance_controller.cpp
altex111: In the joint trajectory controller it was created in this order, thats why i did in this way.

kuka_controllers/src/joint_impedance_controller.cpp
altex111: ok

kuka_rox_hardware_interface/src/robot_manager_node.cpp
altex111: [Result]: control_mode_map refill in callbacks

kuka_rox_hardware_interface/include/kuka_rox_hw_interface/robot_manager_node.hpp
```diff
@@ -54,17 +54,30 @@ class RobotManagerNode : public kroshu_ros2_core::ROS2BaseLCNode
...
   rclcpp::CallbackGroup::SharedPtr cbg_;
   std::vector<std::string> controller_names_;
+  std::map<std::string,std::vector<std::string>> control_mode_map_;
 
   std::shared_ptr<rclcpp_lifecycle::LifecyclePublisher<std_msgs::msg::Bool>> is_configured_pub_;
   std_msgs::msg::Bool is_configured_msg_;
 
+
+
+  const std::string POSITION_CONTROL = "position";
```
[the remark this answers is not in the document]
altex111: For now this is not needed

kuka_rox_hardware_interface/src/rox_hardware_interface.cpp
```diff
@@ -112,14 +112,13 @@ export_command_interfaces()
       info_.joints[i].name,
       HW_IF_STIFFNESS,
       &hw_stiffness_[i]);
-  }
 
-  for (size_t i = 0; i < info_.joints.size(); i++) {
```
altex111, line 117: @VX792 @Svastits In this situation is this the right way to correct the mistake, or it should be emplaced like POS, STIFFNESS, DAMPING, POS, STIFFNESS, etc?

Now it is placeing every POS param and after that the STIFFNESS and DAMPING param like STIFFNESS, DAMPING, STIFFNESS, DAMPING, etc.

kuka_controllers/src/joint_impedance_controller.cpp
altex111: This is connected to the changes in rox_hardwer_interface.cpp

kuka_rox_hardware_interface/src/robot_manager_node.cpp
Svastits: I think you do not need the names and map filling here, only the map's find line

VX792 (changes requested):
I've written another batch of comments. Also, it's not yet part of this PR, but I'd consider putting the sched_param struct and the operations on it into the robot_manager_node.

It's not obvious at a glance which thread's priority we're setting, especially if we are to support control for multiple hardware interfaces (e.g., robot arm + mobile platform). In that case, the latter might not get RT priority if the arm component is not activated.

VX792 (approved):
Approved to get this merged ASAP.

Svastits (approved):

PR #46, opened 2023-01-19
Fix freeze in case of errors

Svastits:


kuka_rox_hardware_interface/src/robot_manager_node.cpp
```diff
@@ -126,6 +126,8 @@ RobotManagerNode::on_activate(const rclcpp_lifecycle::State &)
   if (!hw_response || !hw_response->ok) {
     RCLCPP_ERROR(get_logger(), "Could not activate hardware interface");
     // 'unset config' does not exist, safe to return
+    is_configured_msg_.data = false;
```
kovacsge11, line 129: Doesn't it get back to configured state? So it should still remain configured as far as I see.

kovacsge11 (approved):

PR #53, opened 2023-02-08
Get params from xacro file

VX792:
HW interface side of the other PR, allows the user to modify the client ip, controller ip and control mode parameters without having to recompile the code every single time.


kuka_rox_hardware_interface/src/rox_hardware_interface.cpp
```diff
@@ -23,7 +23,7 @@
 #include <string>
 #include <vector>
 
-#include "nanopb-helpers/nanopb_serialization_helper.h"
```
Svastits, line 26: did it not build with the original?

kuka_rox_hardware_interface/include/kuka_rox_hw_interface/rox_hardware_interface.hpp
```diff
@@ -90,16 +90,18 @@ class KukaRoXHardwareInterface : public hardware_interface::SystemInterface
...
 
-  os::core::udp::communication::UDPReplier udp_replier_ = os::core::udp::communication::UDPReplier(
-    os::core::udp::communication::SocketAddress(CLIENT_IP, CLIENT_PORT));
 
   nanopb::kuka::ecs::v1::ControlSignalExternal control_signal_ext_{
     nanopb::kuka::ecs::v1::ControlSignalExternal_init_default};
   nanopb::kuka::ecs::v1::MotionStateExternal motion_state_external_{
     nanopb::kuka::ecs::v1::MotionStateExternal_init_default};
 
+  nanopb::kuka::core::motion::JointPositions start_pos_{
```
Svastits, line 102: I think this is unused now

kuka_rox_hardware_interface/src/rox_hardware_interface.cpp
VX792: Nope, this is actually an issue from last week. I can change this line to make the CI happy temporarily

Svastits (approved):

PR #58, opened 2023-02-23
Fix/combine parameters

Svastits:


kuka_rox_hardware_interface/include/mock/kuka/motion/external/external_control_mode.pb.h
```diff
@@ -82,8 +84,28 @@ inline const std::string& ExternalControlMode_Name(T enum_t_value) {
   static_assert(::std::is_same<T, ExternalControlMode>::value ||
     ::std::is_integral<T>::value,
     "Incorrect type passed to function ExternalControlMode_Name.");
-  return ::PROTOBUF_NAMESPACE_ID::internal::NameOfEnum(
-    ExternalControlMode_descriptor(), enum_t_value);
+  switch(enum_t_value)
+  {
+    case ExternalControlMode::POSITION_CONTROL:
+      control_mode_string = "POSITION_CONTROL";
```
VX792, line 90: Why do we need control modes as strings? As I can see, this function is only used for logging and you can do the same with enums in less lines.

kuka_rox_hardware_interface/src/robot_manager_node.cpp
```diff
@@ -124,9 +156,52 @@ RobotManagerNode::on_cleanup(const rclcpp_lifecycle::State &)
...
+
+    kuka::ecs::v1::CommandState response;
+
+    if (reader->Read(&response)) {
+      switch (static_cast<int>(response.event())) {
+        case kuka::ecs::v1::CommandEvent::STOPPED:
+        case kuka::ecs::v1::CommandEvent::ERROR:
+          RCLCPP_INFO(get_logger(), "External control stopped");
+          terminate_ = true;
+          if (this->get_current_state().label() == "active") {
```
VX792, line 176: Can't we work with this->get_current_state().id() instead? It's a bit better than hardcoding strings imo.

kuka_rox_hardware_interface/src/rox_hardware_interface.cpp
```diff
@@ -149,13 +162,21 @@ CallbackReturn KukaRoXHardwareInterface::on_activate(const rclcpp_lifecycle::Sta
...
+          "control_mode"))).c_str());
 
   if (stub_->OpenControlChannel(
       &context, request,
       &response)
     .error_code() != grpc::StatusCode::OK)
   {
-    return CallbackReturn::ERROR;
+    RCLCPP_ERROR(rclcpp::get_logger("KukaRoXHardwareInterface"), "OpenControlChannel failed");
+    terminate_ = true;
```
VX792, line 178: Is it possible to only spawn the thread if the OpenControlChannel grpc call is successful, at the end of the on_activate function? It might be possible to avoid setting the terminate_ flag here that way and maybe find a less hacky way to restart it if the activation fails.

kuka_rox_hardware_interface/src/robot_manager_node.cpp
Svastits: fully agree :)

kuka_rox_hardware_interface/include/mock/kuka/motion/external/external_control_mode.pb.h
Svastits: Can you convert the enum name to string in a simple way, or what do you mean?
With the real proto lib, the descriptor is used to generate the string, but in mock I found it easier to just hardcode it to be able to build it

kuka_rox_hardware_interface/src/rox_hardware_interface.cpp
Svastits: The ObserControl request must be called before the OpenControlChannel, as otherwise we would miss the first two state changes (in which case the RT loop would not start)

kuka_rox_hardware_interface/include/mock/kuka/motion/external/external_control_mode.pb.h
VX792: No, I meant logging the enums as numbers, without converting them to strings. Mainly because the user already has to configure the control mode as an int. 

Probably not that important to address right now though, and the only thing you win with this is less lines of code.

VX792 (approved):

kuka_rox_hardware_interface/src/robot_manager_node.cpp
Svastits: is there an enum for the id? I think I could not find one and therefore used the string

PR #76, opened 2023-03-22
add sched_param to the correct place

VX792:
Since the sched_setscheduler function is called in the main, the executor is also spinning on RT priority, which is entirely pointless (especially since it's blocking). This PR solves this by calling the function inside the control_loop thread, where the meaningful stuff happens.


Svastits (approved):

Svastits:
Ok after green CI

Svastits (approved):

PR #80, opened 2023-05-23
Feature/controller for control modes

altex111:


altex111:
Comment clearing is nessecery

kuka_rox_hardware_interface/src/robot_manager_node.cpp
```diff
@@ -46,64 +48,60 @@ RobotManagerNode::RobotManagerNode()
...
-      ExternalControlMode::TORQUE_CONTROL,
-      std::vector<std::string>(STANDARD_MODE_IF_SIZE)));
-
-  // TODO(Svastits): change to dynamic parameter after control mode changes are supported
-  this->registerStaticParameter<int>(
+  control_mode_pub_ = this->create_publisher<std_msgs::msg::UInt32>(
+    "control_mode", rclcpp::SystemDefaultsQoS()
+  );
+
+  control_mode_change_finished_ = false;
```
VX792, line 55: initialize this either in the constructor's initializer list or in the class definition

VX792 (changes requested):

kuka_rox_hardware_interface/src/robot_manager_node.cpp
altex111: done

kuka_rox_hardware_interface/src/robot_manager_node.cpp
```diff
@@ -46,64 +48,60 @@ RobotManagerNode::RobotManagerNode()
...
+  // Register parameters
+  this->registerParameter<int>(
     "control_mode", static_cast<int>(ExternalControlMode::POSITION_CONTROL),
-    kroshu_ros2_core::ParameterSetAccessRights{true, false,
-      false, false, false}, [this](int control_mode) {
-      return this->onControlModeChangeRequest(control_mode);
+    kroshu_ros2_core::ParameterSetAccessRights{true, true,
+      true, false, false}, [this](int control_mode) {
+      // TODO (komaromi):
+      if (control_mode != 2 && control_mode != 4 && control_mode != 6) {
```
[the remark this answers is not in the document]
altex111: The function itself does not have condition for not implemented control modes, because it was created in mind that it can work with every control mode we defined

Also I forgot to write the todo: It has to be removed once control modes are supported

kuka_rox_hardware_interface/src/robot_manager_node.cpp
```diff
@@ -303,24 +321,98 @@ RobotManagerNode::on_deactivate(const rclcpp_lifecycle::State &)
...
     RCLCPP_INFO(
-      get_logger(), "Control mode changed to %s", ExternalControlMode_Name(control_mode).c_str());
+      get_logger(), "Successfully changed control mode to %s", ExternalControlMode_Name(
+        control_mode).c_str());
     return true;
-  } else {
-    RCLCPP_WARN(
-      get_logger(), "Could not change control mode, %s is not a valid control mode",
-      ExternalControlMode_Name(control_mode).c_str());
+  } catch (const std::out_of_range & e) {
```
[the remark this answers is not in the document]
altex111: Modified to catch only one

kuka_rox_hardware_interface/src/robot_manager_node.cpp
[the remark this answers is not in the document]
VX792: The reason I'd handle it inside the function is that you won't have to guard against this outside the function every single time. This is probably not a big deal, but prevents accidental mistakes. It's also not a good practice to introduce contracts whenever you wish to use a function, since the developer / user might not know about those without looking through the source code in detail.

Svastits (changes requested):

Svastits:
Please remove this line from the hw_interface (L66):
  control_signal_ext_.control_signal.control_mode =
    kuka_motion_external_ExternalControlMode_POSITION_CONTROL;

This was added to make the driver work with the new protos until this is merged

kuka_rox_hardware_interface/src/robot_manager_node.cpp
```diff
@@ -223,67 +210,74 @@ void RobotManagerNode::ObserveControl()
 rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn
 RobotManagerNode::on_activate(const rclcpp_lifecycle::State &)
 {
+  try {
```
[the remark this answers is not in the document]
altex111: This catches if there is any error while the controller handler returns with the controller, if the given control mode is out of the range or unspecified

kuka_rox_hardware_interface/src/robot_manager_node.cpp
```diff
@@ -223,67 +210,74 @@ void RobotManagerNode::ObserveControl()
...
-  auto control_mode = this->get_parameter("control_mode").as_int();
-  auto new_controllers = controller_handler_.Get_activate_deactivate_controllers(control_mode);
+    // Activate RT controller(s)
+    auto controller_request =
+      std::make_shared<SwitchController::Request>();
+    controller_request->strictness = SwitchController::Request::STRICT;
+    controller_request->activate_controllers = new_controllers.first;
+    if (!new_controllers.second.empty()) {
+      controller_request->deactivate_controllers = new_controllers.second;
+      RCLCPP_WARN(get_logger(), "Deactivating controllers while activate state");
```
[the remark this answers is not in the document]
altex111: Edited to activation, this should never happen. but if does, it should deactivate those controllers witch should not work

kuka_rox_hardware_interface/src/robot_manager_node.cpp
```diff
@@ -303,24 +321,98 @@ RobotManagerNode::on_deactivate(const rclcpp_lifecycle::State &)
...
+    control_mode_change_finished_ = false;
+    control_mode_lk.unlock();
+
+    // Deactivate controllers
+    if (get_current_state().id() == lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE) {
+      // The driver is in active state
+
+      // Call request for deactivateing controllers for the new control mode
+      controller_request->deactivate_controllers = controllers.second;
+      controller_request->strictness = SwitchController::Request::STRICT;
```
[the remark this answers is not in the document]
altex111: Decided we need strict

kuka_rox_hardware_interface/include/kuka_rox_hw_interface/robot_manager_node.hpp
```diff
@@ -74,17 +97,19 @@ class RobotManagerNode : public kroshu_ros2_core::ROS2BaseLCNode
...
+
   std::shared_ptr<rclcpp_lifecycle::LifecyclePublisher<std_msgs::msg::Bool>> is_configured_pub_;
   std_msgs::msg::Bool is_configured_msg_;
 
   // There are two kinds of control modes with different number of necessary interfaces to be set:
   //  - in standard modes (position, torque), only the control signal to the used interface (1)
   //  - in impedance modes, the setpoint and the parameters describing the behaviour (2)
-  static constexpr int STANDARD_MODE_IF_SIZE = 1;
-  static constexpr int IMPEDANCE_MODE_IF_SIZE = 2;
+  static constexpr int STANDARD_MODE_IF_SIZE = 3;
```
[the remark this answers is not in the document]
altex111: edeted

altex111:
If review passes, CI blocked with ControllerHandler branch PR

kuka_rox_hardware_interface/src/robot_manager_node.cpp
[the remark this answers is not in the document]
Svastits: I think it is clearer if the catch block is wrapped only around the critical function(s)
In this case only the GetControllersForSwitch can throw an exception (at least that is what you want to catch), so please reduce the range of the try block to only that function

kuka_rox_hardware_interface/src/robot_manager_node.cpp
```diff
@@ -229,50 +239,49 @@ RobotManagerNode::on_activate(const rclcpp_lifecycle::State &)
...
+
+    // Activate RT controller(s)
+    auto controller_request =
+      std::make_shared<SwitchController::Request>();
+    controller_request->strictness = SwitchController::Request::STRICT;
+    controller_request->activate_controllers = new_controllers.first;
+    if (!new_controllers.second.empty()) {
+      // This shuld never happen
+      controller_request->deactivate_controllers = new_controllers.second;
+      RCLCPP_WARN(get_logger(), "Deactivating controllers while activating");
```
Svastits, line 256: This log message still does not say anything to a user
"Controller handler state is improper, as active controller list is not empty before activation" would be better
This should indeed never happen, but as rollback is not yet perfectly implemented, I would also make the severity ERROR

kuka_rox_hardware_interface/src/rox_hardware_interface.cpp
```diff
@@ -252,8 +256,8 @@ return_type KukaRoXHardwareInterface::read(
     return return_type::OK;
   }
 
-  if (udp_replier_->ReceiveRequestOrTimeout(receive_timeout_) ==
-    UDPSocket::ErrorCode::kSuccess)
+  if (udp_replier_->ReceiveRequest() ==
```
VX792, line 259: Keep in mind that this only works with a single hardware interface. If we're also adding a mobile platform, then the core control loop could easily get blocked if we aren't receiving a reply from the ECS running on the robot arm's KRC.

kuka_rox_hardware_interface/src/robot_manager_node.cpp
```diff
@@ -309,21 +319,112 @@ RobotManagerNode::on_deactivate(const rclcpp_lifecycle::State &)
...
+
+    if (get_current_state().id() == lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE) {
+      // The driver is in active state
+
+#ifdef NON_MOCK_SETUP
+      // Wait for ObserControl to approve tha robot succefully changed control mode
+      std::unique_lock<std::mutex> control_mode_lk(this->control_mode_cv_m_);
+
+      if (!this->control_mode_cv_.wait_for(
+          control_mode_lk, std::chrono::seconds(3), [this]() {
```
VX792, line 386: Should this be configurable?

VX792 (changes requested):

kuka_rox_hardware_interface/src/mock/os-core-udp-communication/udp_replier_mock.cpp
```diff
@@ -6,46 +6,38 @@
...
-
-std::pair<const uint8_t *, size_t> UDPReplier::GetRequestMessage() const
+Socket::ErrorCode Replier::SendReply(uint8_t * reply_msg_data, size_t reply_msg_size)
 {
-  return {nullptr, 0};
+  // only a mock
 }
 
+std::pair<const uint8_t *, size_t> Replier::GetRequestMessage() const
+{
```
VX792, line 40: Even though this is "only a mock",  missing return statements are undefined behavior. You should return some garbage value here.

kuka_rox_hardware_interface/src/mock/os-core-udp-communication/udp_replier_mock.cpp
```diff
@@ -6,46 +6,38 @@
...
+Socket::ErrorCode Replier::ReceiveRequestOrTimeout(std::chrono::microseconds recv_timeout)
 {
   // only a mock
-  return UDPSocket::ErrorCode::kSuccess;
 }
-
-std::pair<const uint8_t *, size_t> UDPReplier::GetRequestMessage() const
+Socket::ErrorCode Replier::SendReply(uint8_t * reply_msg_data, size_t reply_msg_size)
 {
-  return {nullptr, 0};
```
VX792, line 48: Same here, and in every single function

kuka_rox_hardware_interface/src/mock/os-core-udp-communication/udp_socket_mock.cpp
```diff
@@ -67,125 +67,125 @@ size_t SocketAddress::Size() const {return 0;}
...
-int UDPSocket::Connect(const SocketAddress & remote_address)
+int Socket::Connect(const SocketAddress & remote_address)
 {
   return 0;
 }
 
-int UDPSocket::Select(std::chrono::microseconds timeout, bool read)
+int Socket::Select(std::chrono::microseconds timeout, bool read)
 {
   // TODO(kovacsge): check
```
VX792, line 137: what

kuka_rox_hardware_interface/src/robot_manager_node.cpp
```diff
@@ -181,6 +181,14 @@ void RobotManagerNode::ObserveControl()
 
   while (reader->Read(&response)) {
     switch (static_cast<int>(response.event())) {
+      case kuka::ecs::v1::CommandEvent::COMMAND_MODE_SWITCH:
+        {
+          std::lock_guard<std::mutex> lk(control_mode_cv_m_);
+          control_mode_change_finished_ = true;
+        }
+        RCLCPP_INFO(get_logger(), "Command mode switched");
```
VX792, line 189: Switched or is in progress? I think both are valid, because the switch already happened, but when we receive this msg, the new session might not be active. Just be consistent with the other message (since it'd be weird to see a print like "Command mode switched" before the "Command mode switch in progress message")

kuka_rox_hardware_interface/src/robot_manager_node.cpp
```diff
@@ -309,21 +319,112 @@ RobotManagerNode::on_deactivate(const rclcpp_lifecycle::State &)
...
   RCLCPP_INFO(get_logger(), "Successfully stopped controllers");
   return SUCCESS;
 }
 
 bool RobotManagerNode::onControlModeChangeRequest(int control_mode)
 {
-  if (control_mode_map_.find(control_mode) != control_mode_map_.end()) {
+  try {
+    RCLCPP_INFO(get_logger(), "Control mode change process has started");
+    if (control_mode == kroshu_ros2_core::ControllerHandler::CARTESIAN_POSITION_CONTROL_MODE ||
```
VX792, line 332: what if I'm changing to the control mode with the enum value 456? I think we should also handle that case

kuka_rox_hardware_interface/src/robot_manager_node.cpp
```diff
@@ -309,21 +319,112 @@ RobotManagerNode::on_deactivate(const rclcpp_lifecycle::State &)
...
+        this->on_deactivate(get_current_state());
+        return false;
+      }
+      control_mode_change_finished_ = false;
+      control_mode_lk.unlock();
+#endif
+
+      // Deactivate controllers
+
+      // Call request for deactivateing controllers for the new control mode
```
VX792, line 403: typo -> deactivating

kuka_rox_hardware_interface/src/robot_manager_node.cpp
```diff
@@ -309,21 +319,112 @@ RobotManagerNode::on_deactivate(const rclcpp_lifecycle::State &)
...
+      }
+      controller_handler_.ApproveControllerActivation();
+    }
+
+    // Publish the control mode to controller handler
+    auto message = std_msgs::msg::UInt32();
+    message.data = control_mode;
+    control_mode_pub_->publish(message);
+
+    if (get_current_state().id() == lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE) {
```
VX792, line 378: why are we checking for active state twice?

kuka_rox_hardware_interface/src/robot_manager_node.cpp
```diff
@@ -309,21 +319,112 @@ RobotManagerNode::on_deactivate(const rclcpp_lifecycle::State &)
...
 
   RCLCPP_INFO(get_logger(), "Successfully stopped controllers");
   return SUCCESS;
 }
 
 bool RobotManagerNode::onControlModeChangeRequest(int control_mode)
 {
-  if (control_mode_map_.find(control_mode) != control_mode_map_.end()) {
+  try {
+    RCLCPP_INFO(get_logger(), "Control mode change process has started");
```
VX792, line 331: Getting to this point doesn't actually mean that the process has started, only the ECS knows that it did. As far as I know here we're waiting for the actual control mode change to occur on the server side, and then change the controllers.

The logs should reflect that as currently it's a bit confusing

VX792 (changes requested):

kuka_rox_hardware_interface/src/robot_manager_node.cpp
```diff
@@ -309,21 +319,112 @@ RobotManagerNode::on_deactivate(const rclcpp_lifecycle::State &)
...
+    if (control_mode == kroshu_ros2_core::ControllerHandler::CARTESIAN_POSITION_CONTROL_MODE ||
+      control_mode == kroshu_ros2_core::ControllerHandler::CARTESIAN_IMPEDANCE_CONTROL_MODE ||
+      control_mode == kroshu_ros2_core::ControllerHandler::WRENCH_CONTROL_MODE)
+    {
+      RCLCPP_ERROR(get_logger(), "Tried to change a not implemented control mode");
+      // TODO(Svastits): this can be removed if rollback is implemented properly
+      this->on_deactivate(get_current_state());
```
altex111, line 338: @Svastits what do you think, Is this needed here?

kuka_rox_hardware_interface/src/robot_manager_node.cpp
altex111: The controller handler catch that

kuka_rox_hardware_interface/src/robot_manager_node.cpp
altex111: maybe, what are you suggesting?
@Svastits whats your opinion?

kuka_rox_hardware_interface/src/robot_manager_node.cpp
VX792: Sorry, I'm not sure I understand that correctly. Can the get_current_state().id() at line 351 return a different value than the one at line 383? If so, this might be okay, but if the answer is no, then the second check is redundant.

altex111: modified: The current state check called only once

kuka_rox_hardware_interface/src/robot_manager_node.cpp
altex111: created a variable for it for now, later if needed we can add a parameter

kuka_rox_hardware_interface/src/rox_hardware_interface.cpp
altex111: @Svastits @VX792 I believe if this is resolved and everything tested again the PR should be ready.

Svastits: I have reverted to the original version, after testing it non-mock, we could merge this :)

Svastits: Original version with timeout is working

kuka_rox_hardware_interface/src/mock/os-core-udp-communication/udp_replier_mock.cpp
```diff
@@ -6,46 +6,42 @@
...
-
-UDPSocket::ErrorCode UDPReplier::Setup()
+Replier::Replier(const SocketAddress & local_address)
 {
   // only a mock
-  return UDPSocket::ErrorCode::kSuccess;
 }
-
-void UDPReplier::Reset()
+Socket::ErrorCode Replier::Setup()
```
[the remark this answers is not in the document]
Svastits: what do you mean?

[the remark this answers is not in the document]
VX792: This should return something

PR #84, opened 2023-05-26
Support for iisy11 in startup.launch and moveit planning example

Svastits:


kuka_rox_hardware_interface/src/robot_manager_node.cpp
```diff
@@ -321,6 +327,12 @@ bool RobotManagerNode::onControlModeChangeRequest(int control_mode)
     return false;
   }
 }
+
+bool RobotManagerNode::onRobotModelChangeRequest(std::string robot_model)
```
VX792, line 331: lgtm, but why does this have to return a bool?

VX792 (approved):

kuka_rox_hardware_interface/src/robot_manager_node.cpp
Svastits: this is the callback for a parameter change, so yes

PR #92, opened 2023-06-20
Fix/initial control mode

altex111:
The problem was solved with a premature fix.
The decision was to resolve the issue with a more finetuned solution later


Svastits:
I can only write a comment here for some reason...
If you configure, activate, deactivate and activate again, the control mode handler will not be active IMO
You should separate it from the controller_handler and simply activate in configure transition and deactivate in cleanup

PR #99, opened 2023-08-04
Feature/rsi driver

Svastits:


kuka_rsi_hw_interface/src/rsi_command_handler.cpp
```diff
@@ -0,0 +1,325 @@
...
+        auto i = sprintf(err_buf_, "In \"");
+        std::snprintf(err_buf_ + i, node_name.length_ + 1, "%s", node_name.data_ptr_);
+        i += node_name.length_;
+        std::sprintf(err_buf_ + i, "\" param syntax error found");
+        throw std::logic_error(err_buf_);
+      }
+    } else {
+      // node base has childs
+      for (auto && child : element.GetChilds()) {
+        decodeNode(child, buffer, buffer_it, buffer_size);
```
VX792, line 226: This looks a bit dangerous. This is a pretty long function with a recursive call in the middle. So I assume that you want to do a depth first search here, pass the iterator around, process the current element and then return to the parent.

1) How do you ensure that the buffer won't overflow?

2) Are exceptions thrown here supposed to be caught or is aborting the desired behavior?

3) Can you provide any clear exit conditions at the beginning for the recursive function? Or is there nothing we can do here?

4) Also, can you somehow make the recursive function happen at the end?

4) I assume this will be called multiple times on the root node. Since we are so real-time we're not even allocating dynamically because it's evil check out https://en.wikipedia.org/wiki/Memoization for huge potential performance improvements. For example, this is also how you ensure how a recursive Fibonacci function doesn't hang for bigger inputs.

You can also use a queue and store the nodes here to avoid recursive calls, I'm fine with this solution but you should be able to reason about the correctness of it.

kuka_rsi_hw_interface/src/rsi_command_handler.cpp
```diff
@@ -0,0 +1,325 @@
...
+      std::sprintf(
+        err_buf_,
+        "Syntax error, while node end in the %s node",
+        element.GetName().c_str());
+      throw std::logic_error(err_buf_);
+    }
+  }
+}
+
+void RSICommandHandler::encodeNode(
```
VX792, line 262: rename to encodeNodes

VX792 (changes requested):
I've left a few comments. There are multiple occurrences of some issues so please address each of those.

kuka_rsi_hw_interface/src/rsi_control_node.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+  auto executor = std::make_shared<rclcpp::executors::MultiThreadedExecutor>();
+  auto controller_manager = std::make_shared<controller_manager::ControllerManager>(
+    executor,
+    "controller_manager");
+
+  std::thread control_loop([controller_manager]() {
+      const rclcpp::Duration dt =
+      rclcpp::Duration::from_seconds(1.0 / controller_manager->get_update_rate());
+
```
VX792: maybe add RT scheduling?

kuka_rsi_hw_interface/src/rsi_command_handler.cpp
```diff
@@ -0,0 +1,325 @@
...
+        std::snprintf(err_buf_ + i, node_param.length_ + 1, "%s", node_param.data_ptr_);
+        i += node_param.length_;
+        std::sprintf(
+          err_buf_ + i, ") does not match with any of the %s elements parameters.",
+          element.GetName().c_str());
+        throw std::logic_error(err_buf_);
+      }
+      if (*buffer_it != '=') {
+        auto i = sprintf(err_buf_, "In \"");
+        std::snprintf(err_buf_ + i, node_param.length_ + 1, "%s", node_param.data_ptr_);
```
VX792, line 138: why the + 1 here?

kuka_rsi_hw_interface/src/rsi_command_handler.cpp
Svastits, line 262: I would appreciate a few comments to explain a few things

kuka_rsi_hw_interface/src/rsi_command_handler.cpp
```diff
@@ -0,0 +1,325 @@
...
+    }
+  }
+  if (isBaseLessNode) {
+    idx = sprintf(buffer_it, "/>");
+    if ((size_t)(buffer_it - buffer) + idx >= buffer_size) {
+      throw std::range_error("Out of the buffers range");
+    }
+    buffer_it += idx;
+  } else {
+    buffer_it--;
```
Svastits, line 299: if you added a whitespace before printing a parameter name instead of after the element name, this would not be necessary IMO and would make the code a little cleaner

kuka_rsi_hw_interface/src/rsi_command_handler.cpp
```diff
@@ -0,0 +1,325 @@
...
+  if ((size_t)(buffer_it - buffer) + idx >= buffer_size) {
+    throw std::range_error("Out of the buffers range");
+  }
+  buffer_it += idx;
+  bool isBaseLessNode = false;
+  if (element.GetChilds().size() <= 0) {
+    isBaseLessNode = true;
+  }
+  for (auto && param : element.GetParams()) {
+    if (element.GetName() == param.first) {
```
Svastits, line 276: please add a comment here, that the case where an element has a param with the same name, that is not supported

kuka_rsi_hw_interface/src/rsi_command_handler.cpp
```diff
@@ -0,0 +1,325 @@
...
+    throw std::logic_error(err_buf_);
+  }
+
+  // Validate nodes params
+  size_t numOfParam = 0;
+  bool isBaseLessNode = false;
+  bool isNoMoreParam = false;
+  xml::XMLString node_param;
+  while (!isNoMoreParam && (size_t)(buffer_it - buffer) < buffer_size) {
+    // fast forward to the next non-space character
```
Svastits, line 114: It is ok for now, but I think this cannot handle more special whitespace characters, like tab, line break etc

kuka_rsi_hw_interface/include/xml_handler/xml_element.hpp
```diff
@@ -0,0 +1,115 @@
...
+
+public:
+  XMLElement(const std::string & name)
+  : name_(name) {}
+  XMLElement() = default;
+  ~XMLElement() = default;
+
+  inline std::string GetName() const {return name_;}
+  inline void SetName(const std::string & name) {name_ = name;}
+  inline std::vector<XMLElement> GetChilds() const {return childs_;}
```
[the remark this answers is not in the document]
altex111: the first one is const, so when it can be called the const getter of the data structures in rsi command handler

kuka_rsi_hw_interface/src/rsi_command_handler.cpp
```diff
@@ -0,0 +1,325 @@
...
+  // Fast forward to close bracket
+  for (; (size_t)(buffer_it - buffer) < buffer_size && *buffer_it != '>'; buffer_it++) {
+  }
+  if ((size_t)(buffer_it - buffer) >= buffer_size) {
+    throw std::range_error("Out of the buffers range");
+  }
+  buffer_it++;
+
+  // Could not find all parameter, or found too much
+  if (isBaseLessNode && numOfParam != element.GetParams().size()) {
```
Svastits, line 182: why check isBaseLess here?

kuka_rsi_hw_interface/include/xml_handler/xml_element.hpp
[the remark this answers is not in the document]
VX792: keep in mind that the function returning the vector by value might allocate the vector's elements dynamically, since it has to be copied once. So I'd remove that function.

kuka_rsi_hw_interface/include/xml_handler/xml_element.hpp
```diff
@@ -0,0 +1,115 @@
...
+  inline std::vector<XMLElement> & GetChilds() {return childs_;}
+  inline std::map<std::string, XMLParam, std::less<>> & GetParams() {return params_;}
+
+
+  bool CastParam(const XMLString & key, char * & str_ptr);
+  bool IsParamNameValid(XMLString & key, char * & str_ptr);
+  bool IsNameValid(XMLString & key, char * & str_ptr);
+
+  XMLElement & GetElement(const std::string & elementName);
+  const XMLElement & GetElement(const std::string & elementName) const;
```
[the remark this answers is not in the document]
altex111: handler.GetState().GetElement("Out")

The user will call it when accessing the data

kuka_rsi_hw_interface/src/xml_handler/xml_element.cpp
```diff
@@ -0,0 +1,181 @@
...
+  auto param_it = params_.find(key);
+  auto ret_val = param_it != params_.end();
+  if (ret_val) {
+    errno = 0;
+    switch (param_it->second.param_type_) {
+      case XMLType::BOOL:
+        {
+          char * end;
+          auto res = (bool)strtol(str_ptr, &end, 0);
+          if (res == 0 && (errno != 0 || end == str_ptr)) {
```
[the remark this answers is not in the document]
VX792: I know that, but if no conversion can be performed, then strtol returns 0. 
If the conversion is erroneous, then shouldn't we also return false here?

Here you're saying that if res == 0, then return false only if the errno is set or end == str_ptr.

kuka_rsi_hw_interface/src/xml_handler/xml_element.cpp
```diff
@@ -0,0 +1,181 @@
...
+    child.operator<<(out) << "\n";
+  }
+  return out;
+}
+
+XMLString XMLElement::castXMLString(char * & str_ptr)
+{
+  auto start_ref = str_ptr;
+  for (;
+    *str_ptr != '\0' && *str_ptr != '"' && *str_ptr != ' ' &&
```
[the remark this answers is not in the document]
Svastits: or as I see, this is used for 2 separate cases, one is when the string is between "-s, the other when it is a name and has no " around it, so my previous recommendation is not really valid, two different methods might be a solution for this (but still not in this PR)

kuka_rsi_hw_interface/include/xml_handler/xml_param.hpp
```diff
@@ -0,0 +1,57 @@
...
+
+class XMLParam
+{
+public:
+  XMLType param_type_;
+  std::variant<bool, long, double, XMLString> param_;
+  XMLParam() = default;
+  XMLParam(XMLType type)
+  : param_type_(type) {}
+  size_t ParamSprint(char * & buffer_it, char * const buffer, const size_t buffer_size);
```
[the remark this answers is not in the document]
altex111: okey-okey

kuka_rsi_hw_interface/src/rsi_command_handler.cpp
```diff
@@ -0,0 +1,325 @@
...
+      buffer_it++;
+      if (*buffer_it != '"') {
+        auto i = sprintf(err_buf_, "In \"");
+        std::snprintf(err_buf_ + i, node_param.length_ + 1, "%s", node_param.data_ptr_);
+        i += node_param.length_;
+        std::sprintf(err_buf_ + i, "\" param syntax error found");
+        throw std::logic_error(err_buf_);
+      }
+      buffer_it++;
+      if (!element.CastParam(node_param, buffer_it)) {
```
Svastits, line 152: I would add a comment that incresing buffer_it is not necessary after castParam

kuka_rsi_hw_interface/src/rsi_command_handler.cpp
Svastits: I think I get it, that "value-type" parameters like <abc>1</abc> are parsed later, but you could add a comment here too explaining

kuka_rsi_hw_interface/src/rsi_command_handler.cpp
Svastits: it is +i, needed to increase the buffer pointer because fo the "In \""

kuka_rsi_hw_interface/src/rsi_command_handler.cpp
```diff
@@ -0,0 +1,325 @@
...
+        std::sprintf(err_buf_ + i, "\" param syntax error found");
+        throw std::logic_error(err_buf_);
+      }
+      buffer_it++;
+      numOfParam++;
+    }
+  }
+
+  // Fast forward to close bracket
+  for (; (size_t)(buffer_it - buffer) < buffer_size && *buffer_it != '>'; buffer_it++) {
```
Svastits, line 174: might be done later, but a lot of trash can come between the end tag and the next < tag here
e.g. \<abc a="1"/>hahaha&@{đ<abd> is valid according to the code
I think you were aware of this, but then this should be documented somewhere

kuka_rsi_hw_interface/src/rsi_command_handler.cpp
```diff
@@ -0,0 +1,325 @@
...
+  return true;
+}
+
+void RSICommandHandler::decodeNode(
+  xml::XMLElement & element, char * const buffer, char * & buffer_it,
+  const size_t buffer_size)
+{
+  xml::XMLString node_name(buffer, 0);
+  // fast forward for the first open bracket
+  for (; *buffer_it != '<'; buffer_it++) {
```
Svastits, line 91: might be done later, but a lot of trash can come between the end tag and the next < tag here
e.g. <abc a="1"/>hahaha&@{đ is valid according to the code
I think you were aware of this, but then this should be documented somewhere

kuka_rsi_hw_interface/src/rsi_command_handler.cpp
```diff
@@ -0,0 +1,325 @@
...
```
Svastits, line 91: I think you can just assume that it starts with < and throw if not
or else check buffer overflow

kuka_rsi_hw_interface/src/rsi_command_handler.cpp
```diff
@@ -0,0 +1,325 @@
...
+        std::sprintf(err_buf_ + i, "\" param syntax error found");
+        throw std::logic_error(err_buf_);
+      }
+    } else {
+      // node base has childs
+      for (auto && child : element.Childs()) {
+        decodeNode(child, buffer, buffer_it, buffer_size);
+      }
+    }
+    for (; (size_t)(buffer_it - buffer) < buffer_size && *buffer_it != '<'; buffer_it++) {
```
Svastits, line 229: might be done later, but a lot of trash can come between the end tag and the next < tag here
e.g. hahaha&@{đ is valid according to the code
I think you were aware of this, but then this should be documented somewhere

kuka_rsi_hw_interface/src/rsi_command_handler.cpp
```diff
@@ -0,0 +1,325 @@
...
+        std::sprintf(err_buf_ + i, "\" param syntax error found");
+        throw std::logic_error(err_buf_);
+      }
+      buffer_it++;
+      numOfParam++;
+    }
+  }
+
+  // Fast forward to close bracket
+  for (; (size_t)(buffer_it - buffer) < buffer_size && *buffer_it != '>'; buffer_it++) {
```
Svastits, line 174: you can only exit the previous while loop if you encounter a '>', might be nicer to increase the buffer there, so then this is not necessary
(Other possibility of exiting is buffer overflow, and then you do not want to increase the pointer still, but rather throw an exception

kuka_rsi_hw_interface/src/rsi_command_handler.cpp
VX792: I know, I was referring to the node_param.length_ + 1 instead of node_param.length_

kuka_rsi_hw_interface/src/rsi_command_handler.cpp
```diff
@@ -0,0 +1,325 @@
...
+  // fast forward for the first open bracket
+  for (; *buffer_it != '<'; buffer_it++) {
+    if ((int)(buffer_it - buffer) >= (int)buffer_size) {
+      throw std::range_error("Out of the buffers range");
+    }
+  }
+  buffer_it++;
+  // Validate nodes name
+  if (!element.IsNameValid(node_name, buffer_it)) {
+    auto i = std::sprintf(err_buf_, "The detected name (");
```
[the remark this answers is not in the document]
Svastits: Agreed: short error msgs, no need to check now

kuka_rsi_hw_interface/src/rsi_command_handler.cpp
```diff
@@ -0,0 +1,325 @@
...
+  bool isNoMoreParam = false;
+  xml::XMLString node_param;
+  while (!isNoMoreParam && (size_t)(buffer_it - buffer) < buffer_size) {
+    // fast forward to the next non-space character
+    for (; *buffer_it == ' '; buffer_it++) {
+    }
+    if ((size_t)(buffer_it - buffer) >= buffer_size) {
+      throw std::range_error("Out of the buffers range");
+    }
+    // Check for the nodes hadders end characters
```
Svastits, line 120: node headers

kuka_rsi_hw_interface/src/rsi_command_handler.cpp
Svastits: later, add to issue

kuka_rsi_hw_interface/src/rsi_command_handler.cpp
```diff
@@ -0,0 +1,325 @@
...
+    if (*buffer_it == '/' && *(buffer_it + 1) == '>') {
+      isBaseLessNode = true;
+      isNoMoreParam = true;
+    } else if (*buffer_it == '>') {
+      isNoMoreParam = true;
+    } else {
```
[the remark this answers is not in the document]
Svastits: it is checked, no way to check only once at beginning

kuka_rsi_hw_interface/src/rsi_command_handler.cpp
Svastits: resolved

kuka_rsi_hw_interface/src/rsi_command_handler.cpp
```diff
@@ -0,0 +1,325 @@
...
+      }
+      buffer_it++;
+      if (*buffer_it != '"') {
+        auto i = sprintf(err_buf_, "In \"");
+        std::snprintf(err_buf_ + i, node_param.length_ + 1, "%s", node_param.data_ptr_);
+        i += node_param.length_;
+        std::sprintf(err_buf_ + i, "\" param syntax error found");
+        throw std::logic_error(err_buf_);
+      }
+      buffer_it++;
```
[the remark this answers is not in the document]
Svastits: discussed, checked immediately afterwards

kuka_rsi_hw_interface/src/rsi_command_handler.cpp
```diff
@@ -0,0 +1,325 @@
...
+  const size_t buffer_size)
+{
+  xml::XMLString node_name(buffer, 0);
+  // fast forward for the first open bracket
+  for (; *buffer_it != '<'; buffer_it++) {
+    if ((int)(buffer_it - buffer) >= (int)buffer_size) {
+      throw std::range_error("Out of the buffers range");
+    }
+  }
+  buffer_it++;
```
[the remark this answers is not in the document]
Svastits: discussed, checked immediately afterwards

kuka_rsi_hw_interface/src/rsi_command_handler.cpp
```diff
@@ -0,0 +1,325 @@
...
+    // Check for the nodes hadders end characters
+    if (*buffer_it == '/' && *(buffer_it + 1) == '>') {
+      isBaseLessNode = true;
+      isNoMoreParam = true;
+    } else if (*buffer_it == '>') {
+      isNoMoreParam = true;
+    } else {
+      if (!element.IsParamNameValid(node_param, buffer_it)) {
+        auto i = std::sprintf(err_buf_, "The detected parameter (");
+        std::snprintf(err_buf_ + i, node_param.length_ + 1, "%s", node_param.data_ptr_);
```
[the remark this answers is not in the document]
Svastits: add to issue (later)

kuka_rsi_hw_interface/src/rsi_command_handler.cpp
```diff
@@ -0,0 +1,325 @@
...
+
+  // Node should have childs, but did not found any
+  if (isBaseLessNode && element.GetChilds().size() > 0) {
+    std::sprintf(
+      err_buf_, "%s node should have chields but did not found any",
+      element.GetName().c_str());
+    throw std::logic_error(err_buf_);
+  }
+  // fast forward if
+  for (; *buffer_it == ' '; buffer_it++) {
```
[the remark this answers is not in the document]
Svastits: discussed

kuka_rsi_hw_interface/src/xml_handler/xml_element.cpp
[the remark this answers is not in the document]
Svastits: ok

kuka_rsi_hw_interface/src/xml_handler/xml_element.cpp
```diff
@@ -0,0 +1,181 @@
...
+        break;
+    }
+  }
+  for (auto && child : childs_) {
+    child.operator<<(out) << "\n";
+  }
+  return out;
+}
+
+XMLString XMLElement::castXMLString(char * & str_ptr)
```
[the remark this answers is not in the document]
Svastits: discussed

kuka_rsi_hw_interface/src/xml_handler/xml_element.cpp
```diff
@@ -0,0 +1,181 @@
...
+  }
+}
+
+const XMLElement & XMLElement::GetElement(const std::string & elementName, int depth) const
+{
+  if (elementName == name_) {
+    return *this;
+  }
+  for (auto && child : childs_) {
+    child.GetElement(elementName);
```
Svastits, line 124: you should return here

kuka_rsi_hw_interface/src/xml_handler/xml_element.cpp
```diff
@@ -0,0 +1,181 @@
...
+  return strncmp(name_.c_str(), key.data_ptr_, key.length_) == 0;
+}
+
+XMLElement & XMLElement::Element(const std::string & elementName, int depth)
+{
+  if (elementName == name_) {
+    return *this;
+  }
+  for (auto && child : childs_) {
+    child.Element(elementName, depth + 1);
```
Svastits, line 109: also return

kuka_rsi_hw_interface/src/kuka_hardware_interface.cpp
```diff
@@ -234,6 +235,7 @@ return_type KukaRSIHardwareInterface::read(
     return return_type::ERROR;
   }
   if (!command_handler_.Decode(in_buffer_, UDP_BUFFER_SIZE)) {
+    this->on_deactivate(this->get_state());
```
Svastits, line 238: are these deactivations necessary? I would assume returning failure has the same effect

kuka_rsi_hw_interface/src/rsi_command_handler.cpp
```diff
@@ -56,17 +43,34 @@ RSICommandHandler::RSICommandHandler()
...
   state_data_structure_.Childs().emplace_back(RSol_el);
+  state_data_structure_.Childs().emplace_back(AIPos_el);
+  state_data_structure_.Childs().emplace_back(ASPos_el);
+  state_data_structure_.Childs().emplace_back(Delay_el);
   state_data_structure_.Childs().emplace_back(Ipoc_state_el);
 
   // Command structure
   command_data_structure_.Params()["Type"] = xml::XMLParam(xml::XMLType::STRING);
-  command_data_structure_.SetParam<xml::XMLString>("Type", xml::XMLString("KROSHU"));
+  command_data_structure_.SetParam<xml::XMLString>("Type", xml::XMLString("ImFree"));
```
Svastits, line 73: why roll back to ImFree?

kuka_rsi_hw_interface/include/xml_handler/xml_element.hpp
```diff
@@ -0,0 +1,115 @@
...
+#include <cstring>
+#include <cstdlib>
+#include <iostream>
+#include <functional>
+
+#include "xml_param.hpp"
+
+namespace xml
+{
+inline bool operator<(const XMLString & a, const std::string & b)
```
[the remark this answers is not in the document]
altex111: resolved with size check first then compare

kuka_rsi_hw_interface/include/xml_handler/xml_param.hpp
```diff
@@ -0,0 +1,57 @@
...
+  DOUBLE = 2,
+  STRING = 3
+};
+
+
+class XMLParam
+{
+public:
+  XMLType param_type_;
+  std::variant<bool, long, double, XMLString> param_;
```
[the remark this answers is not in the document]
altex111: The GetParam() changed to GetParamValue
But I did not change the XMLElement::GetParam() func

kuka_rsi_hw_interface/src/rsi_command_handler.cpp
altex111: I did not wanted to commit it, but I did not wanted to bother how the simulator is working

kuka_rsi_hw_interface/src/rsi_command_handler.cpp
altex111: I moved the increment to the while loop

kuka_rsi_hw_interface/src/rsi_command_handler.cpp
altex111: resolved

kuka_rsi_hw_interface/src/rsi_command_handler.cpp
altex111: Still rename it even if the function only encode one node, and calls itself?

kuka_rsi_hw_interface/src/xml_handler/xml_element.cpp
```diff
@@ -0,0 +1,181 @@
...
+bool XMLElement::CastParam(const XMLString & key, char * & str_ptr)
+{
+  auto param_it = params_.find(key);
+  auto ret_val = param_it != params_.end();
+  if (ret_val) {
+    errno = 0;
+    switch (param_it->second.param_type_) {
+      case XMLType::BOOL:
+        {
+          char * end;
```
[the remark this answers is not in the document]
altex111: resolved

kuka_rsi_hw_interface/include/xml_handler/xml_element.hpp
```diff
@@ -0,0 +1,115 @@
...
+  inline std::vector<XMLElement> GetChilds() const {return childs_;}
+  inline std::vector<XMLElement> & GetChilds() {return childs_;}
+  inline std::map<std::string, XMLParam, std::less<>> & GetParams() {return params_;}
+
+
+  bool CastParam(const XMLString & key, char * & str_ptr);
+  bool IsParamNameValid(XMLString & key, char * & str_ptr);
+  bool IsNameValid(XMLString & key, char * & str_ptr);
+
+  XMLElement & GetElement(const std::string & elementName);
```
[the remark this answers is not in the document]
altex111: modified

kuka_rsi_hw_interface/src/rsi_command_handler.cpp
altex111: modified so there is no buffer_it--

kuka_rsi_hw_interface/src/kuka_hardware_interface.cpp
altex111: I see, I encountered an error like this today.
I closed the simulator while the there was an active control and the whole driver died. It was like the control loop was running without any wait.
We should put it into an issue

kuka_rsi_hw_interface/src/rsi_command_handler.cpp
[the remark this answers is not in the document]
altex111: converted errors to stringstream

kuka_rsi_hw_interface/src/rsi_command_handler.cpp
altex111: commented

kuka_rsi_hw_interface/src/rsi_command_handler.cpp
altex111: done

kuka_rsi_hw_interface/src/rsi_command_handler.cpp
altex111: done

kuka_rsi_hw_interface/src/kuka_hardware_interface.cpp
```diff
@@ -152,16 +152,66 @@ CallbackReturn KukaRSIHardwareInterface::on_activate(const rclcpp_lifecycle::Sta
...
+  hw_states_[1] = command_handler_.GetState().GetElement("AIPos")->GetParam<double>("A2") *
+    KukaRSIHardwareInterface::D2R;
+  hw_states_[2] = command_handler_.GetState().GetElement("AIPos")->GetParam<double>("A3") *
+    KukaRSIHardwareInterface::D2R;
+  hw_states_[3] = command_handler_.GetState().GetElement("AIPos")->GetParam<double>("A4") *
+    KukaRSIHardwareInterface::D2R;
+  hw_states_[4] = command_handler_.GetState().GetElement("AIPos")->GetParam<double>("A5") *
+    KukaRSIHardwareInterface::D2R;
+  hw_states_[5] = command_handler_.GetState().GetElement("AIPos")->GetParam<double>("A6") *
+    KukaRSIHardwareInterface::D2R;
```
gavanderhoorn, line 172: You could consider using the [angles](https://index.ros.org/p/angles/) package for these kinds of conversions ([angles::from_degrees(..)](https://docs.ros.org/en/api/angles/html/namespaceangles.html#ac427c1e73e487dcfca841672f719990a), etc).

kuka_rsi_hw_interface/src/xml_handler/xml_element.cpp
```diff
@@ -0,0 +1,151 @@
...
+            return false;
+          }
+          str_ptr = end;
+          param_it->second.param_value_ = res;
+          break;
+        }
+      case XMLType::DOUBLE:
+        {
+          char * end = str_ptr;
+          auto res = strtod(str_ptr, &end);
```
gavanderhoorn, line 60: IIRC, functions like `strtod(..)` use the current locale setting to decode floating point nrs ([doc](https://cplusplus.com/reference/cstdlib/strtod/)).

We've had problems in URDF parsing libraries on German and Dutch systems fi, where commas are used, versus `en_US` and similar locales where dots are used.

kuka_rsi_hw_interface/src/xml_handler/xml_param.cpp
```diff
@@ -0,0 +1,60 @@
...
+#include "xml_handler/xml_param.hpp"
+
+namespace xml
+{
+int XMLParam::PrintParam(char * & buffer_it, int & size_left)
+{
+  int idx = 0;
+  switch (param_type_) {
+    case XMLType::BOOL:
+      idx = snprintf(buffer_it, size_left, "%u", GetParamValue<bool>());
```
gavanderhoorn, line 28: Same comment about locale here I believe.

gavanderhoorn:
Out of curiosity: what's the rationale behind the custom XML parsing/encoding/extraction code?

kuka_rsi_hw_interface/src/xml_handler/xml_element.cpp
altex111: Tanks for the tip, i should try out if RSI is locale-aware, if not I have to search for other solution

kuka_rsi_hw_interface/src/xml_handler/xml_param.cpp
altex111: I believe you are talking about whether the `printf("%f")`  can cause problem with the commas and dots dilemma, here too.
I have just looked on cpp reference. According to that I think it always use dots.
https://en.cppreference.com/w/c/io/fprintf

altex111:
> Out of curiosity: what's the rationale behind the custom XML parsing/encoding/extraction code?
Because of the real-time nature of the communication we needed a real time xml parser, witch do not have any heap allocation in it.
We did not found any witch fitted thees criteria, because they allocate memory while deserialize. So i had to write my own.

gavanderhoorn:
> Because of the real-time nature of the communication we needed a real time xml parser, witch do not have any heap allocation in it. We did not found any witch fitted thees criteria, because they allocate memory while deserialize. So i had to write my own.

Ah, ok.

If you're mostly worried about heap allocations, could using something like TinyXML with [TLSF](http://www.gii.upv.es/tlsf/) be an option? Could save you from maintaining your own parser. It won't do anything for control flow, but at least allocation should be bounded in time that way.

PS: I never really understood why KUKA chose to use XML for the RSI payload, especially because of the real-time requirements for everything. A binary protocol payload would have been much easier.

kuka_rsi_hw_interface/src/rsi_command_handler.cpp
```diff
@@ -0,0 +1,325 @@
...
+        std::snprintf(err_buf_ + i, node_param.length_ + 1, "%s", node_param.data_ptr_);
+        i += node_param.length_;
+        std::sprintf(
+          err_buf_ + i, " param into the %s elements parameter list.",
+          element.GetName().c_str());
+        throw std::logic_error(err_buf_);
+      }
+      if (*buffer_it != '"') {
+        auto i = sprintf(err_buf_, "In \"");
+        std::snprintf(err_buf_ + i, node_param.length_ + 1, "%s", node_param.data_ptr_);
```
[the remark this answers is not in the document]
altex111: resolved

kuka_rsi_hw_interface/src/xml_handler/xml_element.cpp
```diff
@@ -0,0 +1,181 @@
...
+  }
+  return false;
+  // return params_.find(key) != params_.end();
+}
+
+
+bool XMLElement::IsNameValid(XMLString & key, char * & str_ptr)
+{
+  key = castXMLString(str_ptr);
+  std::cout << "detected name: " << key << std::endl;
```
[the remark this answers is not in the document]
altex111: deleted

kuka_rsi_hw_interface/src/xml_handler/xml_element.cpp
```diff
@@ -0,0 +1,181 @@
...
+    return *this;
+  }
+  for (auto && child : childs_) {
+    if (elementName == child.name_) {
+      return child.GetElement(elementName);
+    }
+  }
+  char err_buf[1000];
+  sprintf(err_buf, "%s element not found", elementName.c_str());
+  throw std::logic_error(err_buf);
```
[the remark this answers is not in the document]
altex111: redesigned

kuka_rsi_hw_interface/src/xml_handler/xml_string.cpp
```diff
@@ -0,0 +1,53 @@
...
+  }
+  return strncmp(data_ptr_, rhs.data_ptr_, length_);
+}
+
+bool XMLString::operator==(const std::string & rhs)
+{
+  if (length_ != rhs.length()) {
+    return false;
+  }
+  return strncmp(rhs.c_str(), data_ptr_, length_);
```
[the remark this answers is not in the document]
altex111: Check it again please, The current version is okey?

kuka_rsi_hw_interface/src/xml_handler/xml_string.cpp
```diff
@@ -0,0 +1,53 @@
...
+}
+
+bool XMLString::operator==(const char * & rhs)
```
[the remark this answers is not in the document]
altex111: deleted

kuka_rsi_hw_interface/src/rsi_command_handler.cpp
altex111: added

kuka_rsi_hw_interface/src/xml_handler/xml_element.cpp
```diff
@@ -0,0 +1,181 @@
...
+
+
+std::ostream & XMLElement::operator<<(std::ostream & out) const
+{
+  out << "Element name: " << this->name_ << "\n";
+  for (auto && param : params_) {
+    out << "  - " << param.first << ": ";
+    switch (param.second.param_type_) {
+      case XMLType::BOOL:
+        out << std::get<0>(param.second.param_) << "\n";
```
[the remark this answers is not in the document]
altex111: the function is deleted, else corrected

kuka_rsi_hw_interface/src/xml_handler/xml_element.cpp
gavanderhoorn: You could enforce the C locale inside the driver/this class. That would always use dots AFAIK.

kuka_rsi_hw_interface/src/kuka_hardware_interface.cpp
altex111: thanks for the tip, I corrected it

kuka_rsi_hw_interface/include/xml_handler/xml_element.hpp
```diff
@@ -0,0 +1,115 @@
...
+  const XMLElement & GetElement(const std::string & elementName) const;
+
+  std::ostream & operator<<(std::ostream & out) const;
+
+  // TODO (Komaromi): When cpp20 is in use, use requires so only the types we set can be used
```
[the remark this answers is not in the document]
altex111: resolved

kuka_rsi_hw_interface/src/kuka_hardware_interface.cpp
Svastits: You should deactivate the manager node in case of an error, but here you deactivate the hw if, which is IMO done by simply returning error.

Svastits:
You should remove all code that is not necessary for the reworked driver (e.g. RSIState class)

kuka_rsi_hw_interface/CMakeLists.txt
```diff
@@ -23,35 +23,44 @@ find_package(kroshu_ros2_core REQUIRED)
 find_package(hardware_interface REQUIRED)
 find_package(controller_manager_msgs REQUIRED)
 find_package(pluginlib REQUIRED)
+find_package(angles REQUIRED)
 
 find_package(tinyxml_vendor REQUIRED)
 find_package(TinyXML REQUIRED)
 
 include_directories(include ${TinyXML2_INCLUDE_DIRS})
```
Svastits, line 31: remove this

kuka_rsi_hw_interface/include/xml_handler/xml_string.hpp
```diff
@@ -0,0 +1,48 @@
...
+  : data_ptr_(data_ptr), length_(length) {}
+  explicit inline XMLString(const char * data_ptr)
+  : data_ptr_(data_ptr)
+  {
+    length_ = strnlen(data_ptr, 256);
+    if (length_ == 256) {
+      throw std::range_error("XMLString data length out of range. Range: 0 - 255");
+    }
+  }
+  // explicit XMLString(const std::string & str)
```
Svastits, line 39: remove

kuka_rsi_hw_interface/src/kuka_hardware_interface.cpp
```diff
@@ -152,16 +153,65 @@ CallbackReturn KukaRSIHardwareInterface::on_activate(const rclcpp_lifecycle::Sta
...
+  command_handler_.SetCommandParam<double>("AK", "A6", joint_pos_correction_deg_[5]);
+  command_handler_.SetCommandParam<bool>("Stop", "Stop", stop_flag_);
+  command_handler_.SetCommandParam<int64_t>("IPOC", "IPOC", static_cast<int64_t>(ipoc_));
+
+  auto out_buffer_it = out_buffer_;
+  if (command_handler_.Encode(out_buffer_it, UDP_BUFFER_SIZE) < 0) {
+    RCLCPP_ERROR(rclcpp::get_logger("KukaRSIHardwareInterface"), "Encode Failed");
+    return CallbackReturn::FAILURE;
+  }
+  // out_buffer_ = RSICommand(joint_pos_correction_deg_, ipoc_, stop_flag_).xml_doc;
```
Svastits, line 214: remove

kuka_rsi_hw_interface/src/rsi_command_handler.cpp
```diff
@@ -0,0 +1,371 @@
...
+    err_ss << "The detected start node name \"" << node_name <<
+      "\" does not match the elements name " <<
+      element.GetName() << ".";
+    throw std::logic_error(err_ss.str());
+  }
+
+  // Validate nodes params
+  size_t num_of_param = 0;
+  bool is_param_only_node = false;
+  bool isNoMoreParam = false;
```
Svastits, line 142: could you also change this to snake_case?

kuka_rsi_hw_interface/src/xml_handler/xml_element.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+bool XMLElement::IsParamNameValid(XMLString & key, char * & str_ptr)
+{
+  key = castXMLString(str_ptr);
+  auto param_it = params_.find(key);
+  if (param_it != params_.end()) {
+    return true;
+  }
+  return false;
+  // return params_.find(key) != params_.end();
```
Svastits, line 85: remove

kuka_rsi_hw_interface/include/xml_handler/xml_element.hpp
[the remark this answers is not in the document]
Svastits: I think you should return const &, as @VX792 has suggested

kuka_rsi_hw_interface/src/kuka_hardware_interface.cpp
```diff
@@ -152,16 +153,65 @@ CallbackReturn KukaRSIHardwareInterface::on_activate(const rclcpp_lifecycle::Sta
...
+    command_handler_.GetState().GetElement("ASPos")->GetParam<double>("A5"));
+  initial_joint_pos_[5] = angles::from_degrees(
+    command_handler_.GetState().GetElement("ASPos")->GetParam<double>("A6"));
+
+  // Ipoc data
+  ipoc_ = command_handler_.GetState().GetElement("IPOC")->GetParam<int64_t>("IPOC");
 
   for (size_t i = 0; i < info_.joints.size(); ++i) {
-    hw_states_[i] = rsi_state_.positions[i] * KukaRSIHardwareInterface::D2R;
+    // hw_states_[i] = rsi_state_.positions[i] * KukaRSIHardwareInterface::D2R;
```
Svastits, line 193: remove

kuka_rsi_hw_interface/src/kuka_hardware_interface.cpp
```diff
@@ -152,16 +153,65 @@ CallbackReturn KukaRSIHardwareInterface::on_activate(const rclcpp_lifecycle::Sta
...
     hw_commands_[i] = hw_states_[i];
-    initial_joint_pos_[i] = rsi_state_.initial_positions[i] * KukaRSIHardwareInterface::D2R;
+    // initial_joint_pos_[i] = rsi_state_.initial_positions[i] * KukaRSIHardwareInterface::D2R;
```
Svastits, line 195: remove

kuka_rsi_hw_interface/src/kuka_hardware_interface.cpp
```diff
@@ -152,16 +153,65 @@ CallbackReturn KukaRSIHardwareInterface::on_activate(const rclcpp_lifecycle::Sta
...
   }
-  ipoc_ = rsi_state_.ipoc;
-
-  out_buffer_ = RSICommand(joint_pos_correction_deg_, ipoc_, stop_flag_).xml_doc;
+  // ipoc_ = rsi_state_.ipoc;
```
Svastits, line 197: remove

kuka_rsi_hw_interface/src/kuka_hardware_interface.cpp
```diff
@@ -193,12 +243,28 @@ return_type KukaRSIHardwareInterface::read(
...
-
-  for (std::size_t i = 0; i < info_.joints.size(); ++i) {
-    hw_states_[i] = rsi_state_.positions[i] * KukaRSIHardwareInterface::D2R;
+  if (!command_handler_.Decode(in_buffer_, UDP_BUFFER_SIZE)) {
+    this->on_deactivate(this->get_state());
+    return return_type::ERROR;
   }
-  ipoc_ = rsi_state_.ipoc;
+
+  // for (std::size_t i = 0; i < info_.joints.size(); ++i) {
```
Svastits, line 251: remove

kuka_rsi_hw_interface/src/kuka_hardware_interface.cpp
```diff
@@ -193,12 +243,28 @@ return_type KukaRSIHardwareInterface::read(
...
+    command_handler_.GetState().GetElement("AIPos")->GetParam<double>("A2"));
+  hw_states_[2] = angles::from_degrees(
+    command_handler_.GetState().GetElement("AIPos")->GetParam<double>("A3"));
+  hw_states_[3] = angles::from_degrees(
+    command_handler_.GetState().GetElement("AIPos")->GetParam<double>("A4"));
+  hw_states_[4] = angles::from_degrees(
+    command_handler_.GetState().GetElement("AIPos")->GetParam<double>("A5"));
+  hw_states_[5] = angles::from_degrees(
+    command_handler_.GetState().GetElement("AIPos")->GetParam<double>("A6"));
+  // ipoc_ = rsi_state_.ipoc;
```
Svastits, line 266: remove

kuka_rsi_hw_interface/src/kuka_hardware_interface.cpp
```diff
@@ -217,15 +283,28 @@ return_type KukaRSIHardwareInterface::write(
...
+  command_handler_.SetCommandParam<double>("AK", "A1", joint_pos_correction_deg_[0]);
+  command_handler_.SetCommandParam<double>("AK", "A2", joint_pos_correction_deg_[1]);
+  command_handler_.SetCommandParam<double>("AK", "A3", joint_pos_correction_deg_[2]);
+  command_handler_.SetCommandParam<double>("AK", "A4", joint_pos_correction_deg_[3]);
+  command_handler_.SetCommandParam<double>("AK", "A5", joint_pos_correction_deg_[4]);
+  command_handler_.SetCommandParam<double>("AK", "A6", joint_pos_correction_deg_[5]);
+  command_handler_.SetCommandParam<bool>("Stop", "Stop", stop_flag_);
+  command_handler_.SetCommandParam<int64_t>("IPOC", "IPOC", static_cast<int64_t>(ipoc_));
+
+  // out_buffer_ = RSICommand(joint_pos_correction_deg_, ipoc_, stop_flag_).xml_doc;
```
Svastits, line 298: remove

kuka_rsi_hw_interface/CMakeLists.txt
altex111: why?

kuka_rsi_hw_interface/include/kuka_rsi_hw_interface/rsi_command_handler.hpp
```diff
@@ -0,0 +1,90 @@
+/*********************************************************************
+* Software License Agreement (BSD License)
```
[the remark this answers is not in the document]
altex111: done

kuka_rsi_hw_interface/src/rsi_command_handler.cpp
altex111: done

kuka_rsi_hw_interface/src/kuka_hardware_interface.cpp
Svastits: are you sure this is removed?

kuka_rsi_hw_interface/CMakeLists.txt
Svastits: in the new driver, tinyxml is not used (find_package is also not necessary

Svastits: I meant only the tinyxml include dirs

VX792 (dismissed):

kuka_rsi_hw_interface/src/kuka_hardware_interface.cpp
altex111: I deleted it and resolved this comment

kuka_rsi_hw_interface/CMakeLists.txt
altex111: I thought it was pointing to the find package angles

kuka_rsi_hw_interface/src/kuka_hardware_interface.cpp
```diff
@@ -217,15 +276,28 @@ return_type KukaRSIHardwareInterface::write(
...
+  command_handler_.SetCommandParam<double>("AK", "A3", joint_pos_correction_deg_[2]);
+  command_handler_.SetCommandParam<double>("AK", "A4", joint_pos_correction_deg_[3]);
+  command_handler_.SetCommandParam<double>("AK", "A5", joint_pos_correction_deg_[4]);
+  command_handler_.SetCommandParam<double>("AK", "A6", joint_pos_correction_deg_[5]);
+  command_handler_.SetCommandParam<bool>("Stop", "Stop", stop_flag_);
+  command_handler_.SetCommandParam<int64_t>("IPOC", "IPOC", static_cast<int64_t>(ipoc_));
+
+  auto out_buffer_it = out_buffer_;
+  if (command_handler_.Encode(out_buffer_it, UDP_BUFFER_SIZE) < 0) {
+    this->on_deactivate(this->get_state());
```
Svastits, line 294: another unnecessary deactivate

kuka_rsi_hw_interface/src/xml_handler/xml_element.cpp
Svastits: The German documentation of RSI also uses '.'-s, so I would assume RSI is locale-independent

gavanderhoorn: Right. But what if the PC uses a German locale? IIUC (from the documentation), that would make `strtod(..)` expect `,`, not dots.

altex111: I'm currently makeing the driver to use c locale in these functions, as you suggested earlier

Svastits: According to the setlocale [documentation](https://en.cppreference.com/w/cpp/locale/setlocale), the system always uses the C locale by default:
During program startup, the equivalent of std::setlocale([LC_ALL](http://en.cppreference.com/w/cpp/locale/LC_categories), "C"); is executed before any user code is run.


Therefore IMO it would not be needed to set the locale from the code

gavanderhoorn: I'm not saying you're wrong, but just wanted to link to the issues I mentioned earlier in the URDF parser. See https://github.com/ros/urdfdom_headers/issues/41 fi and the linked issues there.

Svastits: Is it possible that ROS itself set the locale to be the system locale? in our experiments we did not use ROS, just a simple C++ code

gavanderhoorn: I'm not aware of any code in ROS 2 which explicitly sets any locale.

(but then again, ROS is rather large, and I can't claim to have seen everything)

gavanderhoorn:
@Svastits: just so we understand correctly: RSI support is still planned, but the implementation strategy will be different?

Does #144 already work to add RSI support?

Svastits:
Exactly, we still plan to improve RSI support: we will create a non-ROS2 client library/SDK, that we will use in the ROS2 driver. 
#144 is only slightly related to this effort: the SDK is planned to have the same interface as the current [`kuka-external-control-sdk`](https://github.com/kroshu/kuka-external-control-sdk), which currently only support iiQKA cobots. Implementation of the RSI improvements is already ongoig, I can share more details in email :)

gavanderhoorn:
> Implementation of the RSI improvements is already ongoig, I can share more details in email

I'd be interested to learn more about this.

Especially about the current state of the RSI integration.

PR #110, opened 2023-11-10
Adapt to renamings

Svastits:


Svastits:
@Groebehavn, @VX792 I will merge this, as the current master is not compatible with the other repos
Feel free to comment issues afterwards, which I will fix in a later PR

PR #116, opened 2023-11-16
Installation guide

fazakaserik:
Added installation guide to README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,4 +7,37 @@ Github CI | SonarCloud
 ------------| ---------------
 [![Build Status](https://github.com/kroshu//kuka_drivers/workflows/CI/badge.svg?branch=master)](https://github.com/kroshu/ros2_kuka_sunrise_fri_driver/actions) | [![Quality Gate Status](https://sonarcloud.io/api/project_badges/measure?project=kroshu_kuka_drivers&metric=alert_status)](https://sonarcloud.io/dashboard?id=kroshu_kuka_drivers)
 
+# Installation
+
+Create ROS2 workspace (if already not created).
+```bash
+mkdir -p ~/ros2_ws/src
+```
+
+Clone KUKA ROS2 repos.
+```bash
+cd ~/ros2_ws/src
+git clone https://github.com/kroshu/kuka_drivers.git
+vcs import < kuka_drivers/upstream.repos
+```
+
+Get ROS2 dependencies.
+```bash
+cd ~/ros2_ws
+sudo apt update && rosdep install -r --from-paths . --ignore-src --rosdistro $ROS_DISTRO -y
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -7,4 +7,37 @@ Github CI | SonarCloud
...
+```bash
+cd ~/ros2_ws/src
+git clone https://github.com/kroshu/kuka_drivers.git
+vcs import < kuka_drivers/upstream.repos
+```
+
+Get ROS2 dependencies.
+```bash
+cd ~/ros2_ws
+sudo apt update && rosdep install -r --from-paths . --ignore-src --rosdistro $ROS_DISTRO -y
```
Svastits, line 27: I would also add the rosdep init before this, as that is needed if someone uses rosdep for the first time

fazakaserik: Makes sense, I forgot it.

Svastits (approved):

README.md
```diff
@@ -1,10 +1,50 @@
...
+```
+
+Install dependencies using `rosdep`.
+```bash
+cd ~/ros2_ws
+sudo apt install python3-rosdep
+sudo rosdep init
+rosdep update
+sudo apt upgrade
+rosdep install -r --from-paths . --ignore-src --rosdistro $ROS_DISTRO -y
```
gavanderhoorn, line 32:
```suggestion
rosdep install --from-paths . --ignore-src --rosdistro $ROS_DISTRO -y
```

It's only a suggestion, but I would not add `-r` to `rosdep install` command lines.

It let's `rosdep` ignore dependencies it either couldn't resolve or install and continue with "the rest".

That may sound like a good thing, but in reality it just confuses users: `rosdep` will print a summary at the end listing dependencies it couldn't install, but in my experience, as long as users "get back the command prompt", they assume the previous command was successful. Or they don't read `rosdep` output at all, because "there's so much of it".

Failing early (and loudly) is better in those cases: without `-r`, `rosdep` will only print an error and then not do anything.

If you added `-r` to avoid `rosdep` stopping because one of your dependencies doesn't have a `rosdep` key (yet), I'd suggest considering to add a `--skip-key="your_key [other_key [..]]" for now.

README.md
```diff
@@ -1,10 +1,50 @@
...
+cd ~/ros2_ws/src
+git clone https://github.com/kroshu/kuka_drivers.git
+vcs import < kuka_drivers/upstream.repos
+```
+
+Install dependencies using `rosdep`.
+```bash
+cd ~/ros2_ws
+sudo apt install python3-rosdep
+sudo rosdep init
```
gavanderhoorn, line 29: if your user has already ran this command earlier, I believe this will print an error.

PR #120, opened 2023-12-06
add codespell and precommit to CI

VX792:


Svastits:
I think it would be better to use this formatting instead of uncrustify, so that should be removed. 
Additionally, most of the linters are currently executed in the test phase of the build, they should be removed from every package

PR #147, opened 2024-03-05
Adapt to SDK API

VX792:


kuka_iiqka_eac_driver/CMakeLists.txt
```diff
@@ -17,8 +17,7 @@ endif()
 # A few (transitive) dependencies are available only in one of the dependencies
```
Svastits, line 17: please also remove this comment, as the line it was related to, was removed

kuka_iiqka_eac_driver/package.xml
```diff
@@ -18,6 +18,7 @@
   <depend>kuka_drivers_core</depend>
   <depend>hardware_interface</depend>
   <depend>pluginlib</depend>
+  <depend>kuka-external-control-sdk</depend>
```
Svastits, line 21: I think we can only add this if the kuka-external-control-sdk is a ros package or has a rosdep key

Svastits: Yes, CI fails because of this

```
  ERROR: the following packages/stacks could not have their rosdep keys resolved
  to system dependencies:
  kuka_iiqka_eac_driver: Cannot locate rosdep definition for [kuka-external-control-sdk]
```

kuka_iiqka_eac_driver/CMakeLists.txt
VX792: done

PR #148, opened 2024-03-08
Activate joint state broadcaster

Svastits:


kuka_iiqka_eac_driver/src/robot_manager_node.cpp
```diff
@@ -210,9 +210,9 @@ RobotManagerNode::on_activate(const rclcpp_lifecycle::State &)
   std::this_thread::sleep_for(std::chrono::milliseconds(20));
 
   // Activate RT controller(s)
-  if (!kuka_drivers_core::changeControllerState(
-        change_controller_state_client_, controller_handler_.GetControllersForMode(control_mode_),
-        {}))
+  auto controllers = controller_handler_.GetControllersForMode(control_mode_);
+  controllers.push_back("joint_state_broadcaster");
```
VX792, line 214: maybe later we can use a constexpr char[] for the controller names that will never change

VX792 (approved):

kuka_iiqka_eac_driver/src/robot_manager_node.cpp
Svastits: good point, I will include it in this PR

VX792 (dismissed):

PR #152, opened 2024-04-08
Added Omnimove external control 

ravirathnam:


omnimove_external_control/src/omnimove_external_control.cpp
```diff
@@ -0,0 +1,357 @@
...
+hardware_interface::CallbackReturn OmnimoveExternalControl::on_configure(
+  const rclcpp_lifecycle::State & previous_state)
+{
+  return SystemInterface::on_configure(previous_state);
+}
+
+hardware_interface::CallbackReturn OmnimoveExternalControl::on_cleanup(
+  const rclcpp_lifecycle::State & previous_state)
+{
+  return SystemInterface::on_cleanup(previous_state);
```
Svastits, line 86:
```suggestion
  return CallbackReturn::SUCCESS;
```

omnimove_external_control/src/omnimove_external_control.cpp
```diff
@@ -0,0 +1,357 @@
...
+  }
+  catch (std::exception & e)
+  {
+    RCLCPP_ERROR(
+      rclcpp::get_logger("OmnimoveExternalControl"), "Failed to accept connection %s", e.what());
+    return CallbackReturn::ERROR;
+  }
+  RCLCPP_ERROR(rclcpp::get_logger("OmnimoveExternalControl"), "Accepted connection");
+
+  return SystemInterface::on_activate(previous_state);
```
Svastits, line 144:
```suggestion
  return CallbackReturn::SUCCESS;
```

omnimove_external_control/ConfigExtras.cmake
```diff
@@ -0,0 +1 @@
+find_package(Boost REQUIRED COMPONENTS chrono)
```
Svastits, line 1: is there a reason for not calling this direct from the partent cmaklists?

omnimove_external_control/config/omnimove_controller.yaml
Svastits, line 1: I think this config is for the robotic arm itself, not for the platform, so is not necessary here, this package should only contain mobile platform-specific stuff

omnimove_external_control/include/omnimove/external_control_message.hpp
```diff
@@ -0,0 +1,145 @@
...
+class ExternalControlMessage
+{
+private:
+  std::unique_ptr<char[]> msg_buffer_;
+  size_t msg_length_;
+
+protected:
+  const char * header_;
+  const char * message_;
+  const char * crc_;
```
Svastits, line 19: are these used anywhere?

omnimove_external_control/include/omnimove/omnimove_external_control.hpp
```diff
@@ -0,0 +1,52 @@
...
+  boost::chrono::system_clock::time_point last_sent_velocity_time_;
+  std::vector<double> last_sent_velocity_commands_;
+  std::vector<double> position_commands_;
+  std::string agv_type_;
+  std::string protocol_version_;
+  int external_control_port_;
+  boost::asio::io_context io_context_;
+  std::unique_ptr<boost::asio::ip::tcp::acceptor> acceptor_;
+  std::unique_ptr<boost::asio::ip::tcp::socket> client_socket_;
+  boost::circular_buffer<uint8_t> read_buffer_;
```
Svastits, line 48: why do you use a circular buffer? As far as I understood, always only the last message is read, in which case a circ buffer seems to be a little overkill, a simple buffer holding one message might be enough

omnimove_external_control/src/omnimove_external_control.cpp
```diff
@@ -0,0 +1,357 @@
...
+                             .getSerialisedData());
+    }
+    else
+    {
+      if (
+        (boost::chrono::system_clock::now() - last_sent_velocity_time_) >
+        boost::chrono::milliseconds(vel_cmd_timeout_ms_))
+      {
+        client_socket_->send(ExternalControlStopCommand().getSerialisedData());
+      }
```
Svastits, line 327: this block is duplicated a few lines later, I think with a little restructuring these two could be combined outside of the check for the agv type

omnimove_external_control/src/omnimove_external_control.cpp
```diff
@@ -0,0 +1,357 @@
...
+
+  // ExternalControlData received_data(read_buffer_.);
+}
+hardware_interface::return_type OmnimoveExternalControl::read(
+  const rclcpp::Time &, const rclcpp::Duration &)
+{
+  std::array<char, 1024> buffer;
+  if (client_socket_.get() == NULL)
+  {
+    RCLCPP_WARN(rclcpp::get_logger("OmnimoveExternalControl"), "socket is null !!!");
```
Svastits, line 259: I think this is a bigger issue then a warning message, you should return an error

omnimove_external_control/include/omnimove/omnimove_external_control.hpp
```diff
@@ -0,0 +1,52 @@
...
+#include <rcl/logging.h>
+#include <boost/asio.hpp>
+#include <boost/chrono.hpp>
+#include <boost/circular_buffer.hpp>
+#include <hardware_interface/handle.hpp>
+#include <hardware_interface/system_interface.hpp>
+#include <omnimove/external_control_message.hpp>
+namespace omnimove
+{
+class OmnimoveExternalControl : public hardware_interface::SystemInterface
```
Svastits, line 13: class should be rather `KukaOmnimoveHardwareInterface` to align to conventions
file name for header and source should be hardware_interface.h/.cpp
include directory name should match the package name

examples/kuka_rsi_robot_tests/package.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
+  <name>kuka_rsi_robot_tests</name>
```
Svastits, line 4: I think `kuka_mobile_robot_examples` would be better, as this includes examples for mobile platform + robot integration rather than tests

examples/kuka_rsi_robot_tests/kuka_rsi_robot_tests/quantec_test.py
Svastits, line 1: this example is almost the same code as the iontec test, would be better to combined these with some parameters

examples/kuka_rsi_robot_tests/kuka_rsi_robot_tests/quantec_test.py
```diff
@@ -0,0 +1,310 @@
...
+    def action_submit_callback(self, future):
+        self.get_logger().info("action submitted, waiting for action to be done")
+        result_action = future.result().get_result_async()
+        result_action.add_done_callback(self.action_done_callback)
+
+    def action_done_callback(self, future):
+        self.get_logger().info("action finished in callback")
+        self.start_next_movement()
+
+    def start_robot_movements(self):
```
Svastits, line 222: is this used?

omnimove_external_control/CMakeLists.txt
Svastits, line 1: what is the startup sequence for this driver?
all of our other drivers have a robot_manager node with a lifecycle interface, where activation starting the actual real-time communication. It would be nice to also add this component here, as the launch sequence is non-deterministic in timing which could lead to severe issues in my opinion

Svastits (changes requested):

Svastits:
please add a short doc somewhere about setup and launch process

omnimove_external_control/include/omnimove/external_control_message.hpp
```diff
@@ -0,0 +1,145 @@
...
+    uint8_t manual_active_;
+    uint8_t target_height_reached_;
+    uint8_t actual_speed_;
+    uint8_t actual_pos_;
+  };
+
+  PillarData pillar1_;
+  PillarData pillar2_;
+  PillarData pillar3_;
+  PillarData pillar4_;
```
Svastits, line 76: the four pillars could be combined to a single std::array (or std::vector), which would also make copying easier and shorter in some cases

omnimove_external_control/include/omnimove/external_control_message.hpp
```diff
@@ -0,0 +1,145 @@
...
+  uint32_t status_optical_track_guiding_;
+  uint8_t warning_optical_track_guiding_;
+  uint8_t error_optical_track_guiding_;
+  int32_t track_deviation_x_;
+  int32_t track_deviation_y_;
+  int32_t track_deviation_w_;
+  uint8_t tag_number_front_camera_;
+  uint8_t tag_number_rear_camera_;
+  uint8_t tag_number_left_camera_;
+  uint8_t tag_number_right_camera_;
```
Svastits, line 64: As I see, these variables are not used, so they just make it harder to understand the code.
If these are to be used in the future, a controller would also be needed to broadcast the values to the ROS2 world, so at first I would just delete these and readd them later, if the ros2_controller is also ready

omnimove_external_control/src/omnimove_external_control.cpp
```diff
@@ -0,0 +1,357 @@
...
+      velocity_state_[5] = (double)readData.speedPillar4();
+      position_state_[4] = (double)readData.posShield();
+      velocity_state_[6] = (double)readData.speedShield();
+    }
+    else
+    {
+      velocity_state_[0] = (double)readData.speedX();
+      velocity_state_[1] = (double)readData.speedY();
+      velocity_state_[2] = (double)readData.speedW();
+    }
```
Svastits, line 296: It is better practice to use static cast instead of C-style function-like casting
```suggestion
      velocity_state_[0] = static_cast<double>(readData.speedX());
      velocity_state_[1] = static_cast<double>(readData.speedW());
      position_state_[0] = static_cast<double>(readData.posPillar1());
      velocity_state_[2] = static_cast<double>(readData.speedPillar1());
      position_state_[1] = static_cast<double>(readData.posPillar2());
      velocity_state_[3] = static_cast<double>(readData.speedPillar2());
      position_state_[2] = static_cast<double>(readData.posPillar3());
      velocity_state_[4] = static_cast<double>(readData.speedPillar3());
      position_state_[3] = static_cast<double>(readData.posPillar4());
      velocity_state_[5] = static_cast<double>(readData.speedPillar4());
      position_state_[4] = static_cast<double>(readData.posShield());
      velocity_state_[6] = static_cast<double>(readData.speedShield());
    }
    else
    {
      velocity_state_[0] = static_cast<double>(readData.speedX());
      velocity_state_[1] = static_cast<double>(readData.speedY());
      velocity_state_[2] = static_cast<double>(readData.speedW());
    }
```

omnimove_external_control/src/external_control_message.cpp
```diff
@@ -0,0 +1,194 @@
...
+
+namespace omnimove
+{
+
+ExternalControlMessage::ExternalControlMessage(
+  const char * header, const char * message, const int & data_length)
+{
+  size_t header_length = strlen(header);
+  msg_length_ = header_length + data_length + 2;
+  msg_buffer_ = std::make_unique<char[]>(msg_length_);
```
Svastits, line 15: this will be only a problem, when going for real-time, but any allocation on the heap is not allowed in a real-time context (heap allocations should happen in the initialization phase, not in the cyclic part, this constructor is called in every cycle)

examples/kuka_rsi_robot_tests/test/test_copyright.py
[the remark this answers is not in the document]
ravirathnam: removed

omnimove_external_control/package.xml
```diff
@@ -0,0 +1,20 @@
...
+  <version>0.0.0</version>
+  <description>TODO: Package description</description>
+  <maintainer email="osboxes@todo.todo">osboxes</maintainer>
+  <license>TODO: License declaration</license>
+
+  <buildtool_depend>ament_cmake</buildtool_depend>
+  <depend>hardware_interface</depend>
+  <depend>pluginlib</depend>
+
+  <test_depend>ament_lint_auto</test_depend>
```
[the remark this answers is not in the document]
ravirathnam: removed files from test, and removed from package.xml

omnimove_external_control/src/omnimove_external_control.cpp
```diff
@@ -0,0 +1,357 @@
...
+  RCLCPP_INFO(
+    rclcpp::get_logger("OmnimoveExternalControl"), "port of external control server:%d",
+    external_control_port_);
+  return rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn::SUCCESS;
+}
+
+hardware_interface::CallbackReturn OmnimoveExternalControl::on_configure(
+  const rclcpp_lifecycle::State & previous_state)
+{
+  return SystemInterface::on_configure(previous_state);
```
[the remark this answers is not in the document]
ravirathnam: I don't want to assume that, Since i have no further checks of my own, i prefer to call the parent function. If someone smart implements checks on the parent class, i would prefer to have them as well.

omnimove_external_control/src/omnimove_external_control.cpp
ravirathnam: same as earlier, i would like to use the functionality of the parent class. If some functionality is added there, then i get it as well.

omnimove_external_control/src/omnimove_external_control.cpp
ravirathnam: same as the other two comments. What do you think?

omnimove_external_control/include/omnimove/external_control_message.hpp
ravirathnam: removed the unused variables

omnimove_external_control/src/omnimove_external_control.cpp
Svastits: I do not think the base class will be ever extended with checks, as the condition for the transitions (whether the requested transition is valid) is checked before the callback is even called.
However this does not really cause an overhead, so it can remain as it is, if you prefer it this way :)

examples/kuka_rsi_robot_tests/package.xml
ravirathnam: actually, these tests are on the rsi_robot,(only robot movements), not platform movements. The platform tests, i have in the driver package directly

omnimove_external_control/include/omnimove/external_control_message.hpp
ravirathnam: done. combined the four pillars into one

omnimove_external_control/src/omnimove_external_control.cpp
ravirathnam: done

examples/kuka_rsi_robot_tests/package.xml
Svastits: okay sorry, you are right, these are indeed robot tests. 
However I cannot see any RSI-specific commands, so they are rather general tests/examples for kuka robotic arms, are they not? In this case, `kuka_driver_movement_examples`/ `kuka_driver_movement_tests` or something similiar might be more appropriate

examples/kuka_rsi_robot_tests/kuka_rsi_robot_tests/quantec_test.py
Svastits: I think the best would be to use a parameter file with the position values. Evertyhing else could be in one python file that would be valid for all KUKA robots (not only using RSI)

examples/kuka_rsi_robot_tests/package.xml
ravirathnam: yes, you are correct, the tests are not specific to rsi

ravirathnam: Sorry, actually, the test is specific to rsi because, it's startup sequence has the activation of the rsi driver.

examples/kuka_rsi_robot_tests/kuka_rsi_robot_tests/quantec_test.py
ravirathnam: hmmm. Seems like it is not used anymore. will remove it.

omnimove_external_control/CMakeLists.txt
ravirathnam: I use the standard ros robot_manager node. So i didn't need to reinvent my own. I guess it becomes clear with the changes in kuka_robot_descriptions repo.

omnimove_external_control/config/omnimove_controller.yaml
ravirathnam: sorry, file was there in the initial stages, no longer used, has been removed

examples/kuka_rsi_robot_tests/kuka_rsi_robot_tests/quantec_test.py
ravirathnam: movements are fully configured in a parameter file, fully extendible

omnimove_external_control/include/omnimove/omnimove_external_control.hpp
ravirathnam: renamed file. 
Regarding the name to .h. I think it is fine to name cpp header file as hpp. That seems to be convention in ROS

omnimove_external_control/CMakeLists.txt
```diff
@@ -0,0 +1,75 @@
+cmake_minimum_required(VERSION 3.8)
+project(omnimove_external_control)
```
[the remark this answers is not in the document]
ravirathnam: renamed package

examples/kuka_rsi_robot_tests/package.xml
Svastits: But the activation sequence is the same for all KUKA drivers, so I would say it is not RSI specific

kuka_utv3_ec_driver/package.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
+  <name>kuka_utv3_ec_driver</name>
+  <version>0.0.1</version>
+  <description>A ROS2 hardware interface for use with KUKA UTV3 vehicles</description>
+  <maintainer email="ravi.rathnam@kuka.com">Ravi Rathnam</maintainer>
+  <license>Apache 2.0</license>
```
Svastits, line 8: Use [SPDX license identifiers](https://spdx.org/licenses).
```suggestion
  <license>Apache-2.0</license>
```

omnimove_external_control/include/omnimove/omnimove_external_control.hpp
Svastits: But what is the relevance of an outdated message? If you miss a message, that should not be valid in the next tick, you you could simply overwrite it with the fresh one

examples/kuka_rsi_robot_tests/kuka_rsi_robot_tests/quantec_test.py
Svastits: how do you start the example? Would not be a launch file necessary to "give"
 the config files to the node?

omnimove_external_control/CMakeLists.txt
Svastits: What do you mean by the standard robot_manager node? This node was implemented by us and is part of the common non-real time interface for all drivers
The reasoning for needing this addtional node can be found in the [wiki](https://github.com/kroshu/kuka_drivers/wiki#non-real-time-interface)
Please add this robot_manager node to your driver as well, and move the config and launch files to this package too, as is the convention in this repo

Svastits:
Another thing that is missing is the controller for calculating the offset from the starting position that could update the transforms from world to base_link to enable visualization and planning

Svastits:
Closing this due to inactivity

PR #163, opened 2024-04-29
Feature/velocity control pr

krmihaly:
PR for mobile robot cartesian velocity control support

**This will not go to master, as server side is not released**


kuka_iiqka_eac_driver/config/ros2_controller_config.yaml
```diff
@@ -14,3 +14,5 @@ controller_manager:
       type: kuka_controllers/ControlModeHandler
     event_broadcaster:
       type: kuka_controllers/EventBroadcaster
+    twist_controller:
```
VX792, line 17: I think there should also be a separate yaml for the KUKA_MR specific controllers

kuka_iiqka_eac_driver/src/hardware_interface.cpp
```diff
@@ -270,6 +411,19 @@ return_type KukaEACHardwareInterface::write(const rclcpp::Time &, const rclcpp::
   robot_ptr_->GetControlSignal().AddTorqueValues(hw_torque_commands_);
   robot_ptr_->GetControlSignal().AddStiffnessAndDampingValues(
     hw_stiffness_commands_, hw_damping_commands_);
+    
+    hw_twist_commands_vector_[0]=hw_twist_commands_.linear.x;
```
VX792, line 415: Are we sure these copies are needed? I think the controller should do this

kuka_iiqka_eac_driver/include/kuka_iiqka_eac_driver/hardware_interface.hpp
```diff
@@ -34,6 +34,9 @@
 #include "kuka_drivers_core/hardware_event.hpp"
```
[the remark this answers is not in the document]
Svastits: A new launch file should be certainly added, as this way we load a lot of unnecessary controllers
new hwif is also good idea

kuka_iiqka_eac_driver/src/hardware_interface.cpp
```diff
@@ -160,6 +249,18 @@ KukaEACHardwareInterface::export_command_interfaces()
   command_interfaces.emplace_back(
     hardware_interface::CONFIG_PREFIX, hardware_interface::CONTROL_MODE, &hw_control_mode_command_);
 
+  command_interfaces.emplace_back(
```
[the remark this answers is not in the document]
Svastits: agreed, this also points towards a different hwif

kuka_iiqka_eac_driver/src/hardware_interface.cpp
```diff
@@ -235,26 +336,66 @@ return_type KukaEACHardwareInterface::read(const rclcpp::Time &, const rclcpp::D
...
+    }
 
-  // Modify state interface only in read
-  std::lock_guard<std::mutex> lk(event_mutex_);
-  server_state_ = static_cast<double>(last_event_);
+      // Modify state interface only in read
+      std::lock_guard<std::mutex> lk(event_mutex_);
+      server_state_ = static_cast<double>(last_event_);
+
+      tf2::Quaternion q(
```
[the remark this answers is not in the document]
Svastits: Does that make a difference? The `read` will certainly be blocked, so the controller `update` will be delayed the same in each case

kuka_iiqka_eac_driver/src/hardware_interface.cpp
Svastits: Agreed, you currently store the same data twice in different structures, but the command interfaces exported will be a simple vector in all cases (with names making everything straightforward)
so you can just add the vector at the command interface configuration, the `hw_twist_commands_` is not needed

kuka_iiqka_eac_driver/src/hardware_interface.cpp
```diff
@@ -42,7 +42,70 @@ CallbackReturn KukaEACHardwareInterface::on_init(const hardware_interface::Hardw
...
+    }
+    if (joint.state_interfaces[2].name != hardware_interface::HW_IF_Z)
+    {
+      RCLCPP_FATAL(
+        rclcpp::get_logger("KukaEACHardwareInterface"),
+        "expecting 'Z' state interface as third");
+      return CallbackReturn::ERROR;
+    }
+  }
+}
```
Svastits, line 107: these interfaces are not connected to an actual joint, so these check are not necessary (see my [comment ](https://github.com/kroshu/kuka_robot_descriptions/pull/56#discussion_r1588798220)in the robot_descriptions PR)

controllers/twist_controller/src/twist_controller.cpp
```diff
@@ -0,0 +1,129 @@
...
+}
+
+controller_interface::InterfaceConfiguration TwistController::
+command_interface_configuration()
+const
+{
+  controller_interface::InterfaceConfiguration config;
+  config.type = controller_interface::interface_configuration_type::INDIVIDUAL;
+  // config.names.emplace_back("twist commands");
+  config.names.emplace_back("twist/linear/x");
```
Svastits, line 31: you should use the defined constexpressions here. the main reason for defining them is to be able to use them at more places

controllers/twist_controller/src/twist_controller.cpp
```diff
@@ -0,0 +1,129 @@
...
+  return controller_interface::CallbackReturn::SUCCESS;
+}
+
+controller_interface::CallbackReturn
+TwistController::on_deactivate(const rclcpp_lifecycle::State &)
+{
+  return controller_interface::CallbackReturn::SUCCESS;
+}
+
+controller_interface::return_type TwistController::update(
```
Svastits, line 91: this controller has two well-separable tasks:
- forward commands received on a topic to the hwif
- broadcast tf information
I think it would be beneficial to separate these two tasks. (e.g. for robot arms, there is a separate `joint_trajectory_controller` and a `joint_state_broadcaster` that allows exchanging only one of the control and broadcasting logic.)

Another thing to consider is how much can be reused from already existing code. I have looked at the [steering_controllers_library](https://github.com/ros-controls/ros2_controllers/blob/master/steering_controllers_library/doc/userdoc.rst), which seems to convert the `Twist` commands to joint level, so is not appropriate for this task, but you could simply extend the [forward_command_controller](https://github.com/ros-controls/ros2_controllers/blob/master/forward_command_controller/doc/userdoc.rst) for the controlling part and reduce your code considerably.
For the broadcaster part there is no base I know of, so that can remain, however in a separate controller

controllers/twist_controller/src/twist_controller.cpp
```diff
@@ -0,0 +1,129 @@
...
+  config.names.emplace_back("twist/angular/z");
+  return config;
+}
+
+controller_interface::InterfaceConfiguration TwistController::state_interface_configuration()
+const
+{
+  controller_interface::InterfaceConfiguration config;
+  config.type = controller_interface::interface_configuration_type::INDIVIDUAL;
+  config.names.emplace_back("twist/linear/x");
```
Svastits, line 45: these are not used at all, so should not be specified here
a controller should only claim the interfaces it actually uses, it is not a problem, that some interfaces remain unclaimed

kuka_iiqka_eac_driver/src/hardware_interface.cpp
```diff
@@ -132,7 +195,33 @@ std::vector<hardware_interface::StateInterface> KukaEACHardwareInterface::export
 
   state_interfaces.emplace_back(
     hardware_interface::STATE_PREFIX, hardware_interface::SERVER_STATE, &server_state_);
+  state_interfaces.emplace_back(
+      info_.joints[0].name, "x", &hw_twist_state_.linear.x);
```
Svastits, line 199: defined constrexpression should be also used here

kuka_iiqka_eac_driver/src/hardware_interface.cpp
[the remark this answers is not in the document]
Svastits: absolutely true, I only thought about the scheduling of this process

kuka_kmr_iisy_eac_driver/src/hardware_interface.cpp
```diff
@@ -0,0 +1,232 @@
...
+  RCLCPP_DEBUG( rclcpp::get_logger(
+        "KukaRoXHardwareInterface"), "STATE position: %f, %f, %f",
+            hw_pose_state_[0],
+            hw_pose_state_[1],
+            yaw);
+  RCLCPP_DEBUG( rclcpp::get_logger(
+        "KukaRoXHardwareInterface"), "STATE velocity: x:%f, y:%f, theta:%f",
+            hw_twist_state_[0],
+            hw_twist_state_[1],
+            hw_twist_state_[5]);
```
Svastits, line 154: no rox please

controllers/cartesian_pose_broadcaster/include/cartesian_pose_broadcaster.hpp
```diff
@@ -0,0 +1,72 @@
...
+#include <geometry_msgs/msg/transform_stamped.h>
+#include "cartesian_pose_broadcaster_parameters.hpp"
+
+#include "pluginlib/class_list_macros.hpp"
+
+namespace kuka_controllers
+{
+class CartesianPoseBroadcaster : public controller_interface::ControllerInterface
+{
+public:
```
Svastits, line 40: please use the visiblity-control macros, otherwise they are defined in vain
POSE_BROADCASTER_PUBLIC should be used before all public methods, and POSE_BROADCASTER_LOCAL before privates

kuka_iiqka_eac_driver/src/robot_manager_node.cpp
```diff
@@ -270,7 +147,8 @@ bool RobotManagerNode::onControlModeChangeRequest(int control_mode)
   if (
     control_mode != static_cast<int>(kuka_drivers_core::ControlMode::JOINT_POSITION_CONTROL) &&
     control_mode != static_cast<int>(kuka_drivers_core::ControlMode::JOINT_IMPEDANCE_CONTROL) &&
-    control_mode != static_cast<int>(kuka_drivers_core::ControlMode::JOINT_TORQUE_CONTROL))
+    control_mode != static_cast<int>(kuka_drivers_core::ControlMode::JOINT_TORQUE_CONTROL) &&
+    control_mode != static_cast<int>(kuka_drivers_core::ControlMode::CARTESIAN_VELOCITY_CONTROL))
```
Svastits, line 151: I think this is not needed in the iisy robot manager node

kuka_kmr_iisy_eac_driver/config/twist_controller_config.yaml
```diff
@@ -0,0 +1,4 @@
+twist_controller:
+  ros__parameters:
+    joints:
+    - " "
```
Svastits, line 4: should not this joint be named `twist`? The twitst controller's interface names seem to indicate that

kuka_kmr_iisy_eac_driver/include/kuka_kmr_iisy_eac_driver/hardware_interface.hpp
```diff
@@ -0,0 +1,79 @@
...
+#include <memory>
+#include <mutex>
+#include <string>
+#include <thread>
+#include <vector>
+
+#include "hardware_interface/system_interface.hpp"
+#include "pluginlib/class_list_macros.hpp"
+#include "rclcpp/macros.hpp"
+#include "rclcpp/rclcpp.hpp"
```
Svastits, line 29: are these not duplicate (included in the base class)?

kuka_kmr_iisy_eac_driver/launch/startup.launch.py
```diff
@@ -0,0 +1,222 @@
...
+    controller_ip = LaunchConfiguration("controller_ip")
+    client_ip = LaunchConfiguration("client_ip")
+    use_fake_hardware = LaunchConfiguration("use_fake_hardware")
+    ns = LaunchConfiguration("namespace")
+    x = LaunchConfiguration("x")
+    y = LaunchConfiguration("y")
+    z = LaunchConfiguration("z")
+    roll = LaunchConfiguration("roll")
+    pitch = LaunchConfiguration("pitch")
+    yaw = LaunchConfiguration("yaw")
```
Svastits, line 35: I thinks these are not used, remove args

kuka_kmr_iisy_eac_driver/src/robot_manager_node.cpp
```diff
@@ -0,0 +1,213 @@
...
+  terminate_ = false;
+  // Activate hardware interface
+  if (!kuka_drivers_core::changeHardwareState(
+        change_hardware_state_client_, robot_model_, State::PRIMARY_STATE_ACTIVE, 5000))
+  {
+    RCLCPP_ERROR(get_logger(), "Could not activate hardware interface");
+    return FAILURE;
+  }
+
+  // Workaround until controller_manager/jtc bug is fixed:
```
Svastits, line 78: no JTC in this setup, please remove sleep

kuka_kmr_iisy_eac_driver/src/robot_manager_node.cpp
```diff
@@ -0,0 +1,213 @@
...
+  {
+    RCLCPP_ERROR(get_logger(), "Could not stop RT controllers");
+    return ERROR;
+  }
+
+  RCLCPP_INFO(get_logger(), "Successfully stopped controllers");
+  return SUCCESS;
+}
+
+bool MobileRobotManagerNode::onControlModeChangeRequest(int control_mode)
```
Svastits, line 129: this driver only supports 1 control mode, so ideally this parameter would be removed.
But I think you can leave this for now, theoretically there is a possibiliy that a new control mode is added and you should not spend effort on this removal, so this is just a remark

kuka_kmr_iisy_eac_driver/test/test_multi_robot_startup.py
Svastits, line 1: this test is for iisy, please remove

kuka_kmr_iisy_eac_driver/test/config/test1_effort_controller_config.yaml
Svastits, line 1: please delete all config files used only in the multi-robot test (which should be also removed, see my later comment)

kuka_kmr_iisy_eac_driver/CMakeLists.txt
```diff
@@ -0,0 +1,67 @@
...
+  DESTINATION lib/${PROJECT_NAME})
+
+install(DIRECTORY config launch test
+  DESTINATION share/${PROJECT_NAME})
+
+if(BUILD_TESTING)
+  find_package(launch_testing_ament_cmake)
+  add_launch_test(test/test_driver_startup.py)
+  add_launch_test(test/test_driver_activation.py)
+  add_launch_test(test/test_multi_robot_startup.py)
```
Svastits, line 61: no multi-robot test

Svastits (changes requested):
Almost there ;)
Please also fix CI and format here

Svastits:
Closing this, as server side will not be released and relevant comments have been fixed

PR #175, opened 2024-07-26
Migrate master branch to ROS Jazzy

PetoAdam:
Change master branch to use ROS 2 Jazzy, ROS 2 Humble support moves to the "humble" branch.


README.md
```diff
@@ -2,12 +2,16 @@
...
+**Humble** | [`humble`](https://github.com/kroshu/kuka_drivers/tree/humble) | [![Build Status](https://github.com/kroshu//kuka_drivers/actions/workflows/industrial_ci.yml/badge.svg?branch=humble)](https://github.com/kroshu/kuka_drivers/actions) | [![Quality Gate Status](https://sonarcloud.io/api/project_badges/measure?project=kroshu_kuka_drivers&metric=alert_status&branch=humble)](https://sonarcloud.io/dashboard?id=kroshu_kuka_drivers)
 
 # Requirements
-The drivers require a system with ROS installed. It is recommended to use Ubuntu 22.04 with ROS humble. Iron Irwini has breaking changes in the moveit API, thus it is not yet supported.
+The drivers require a system with ROS installed. It is recommended to use Ubuntu 24.04 with ROS Jazzy.
+
+Additionally, there exists a ROS Humble version of the drivers, and its corresponding configuration can be found under the ‘humble’ branch.
```
Svastits, line 13: nicer (AltGr + 7)
```suggestion
Additionally, there exists a ROS Humble version of the drivers, and its corresponding configuration can be found under the `humble` branch.
```

PetoAdam: done

kuka_sunrise_fri_driver/test/test_driver_activation.py
```diff
@@ -71,7 +71,6 @@ def generate_test_description():
 class TestDriverActivation(unittest.TestCase):
     def test_read_stdout(self, proc_output):
         # Check for successful initialization
-        proc_output.assertWaitFor("got segment base", timeout=5)
```
Svastits, line 74: the new log messages in jazzy are e.g.
`[robot_state_publisher-3] [INFO] [1721996934.659519468] [robot_state_publisher]: Robot initialized`
please fix tests to wait for this message

Svastits:
> Change master branch to use ROS 2 Jazzy, ROS 2 Humble support moves to the "humble" branch.

To make this transition seemless, this PR should be merge only after the sources has been updated in the rosdistro repo (I will do it, but humble branches would be necessary for the two kroshu upstream repos as well)

kuka_sunrise_fri_driver/test/test_driver_activation.py
PetoAdam: Done.
Note: I checked the kroshu upstream repos as well, the robot descriptions also use the "got segment" check for ros-in frames:
```
class TestModels(unittest.TestCase):
    def test_read_stdout(self, proc_output):
        # Check for frames defined by ROS-Industrial
        proc_output.assertWaitFor("got segment base", timeout=5)
        proc_output.assertWaitFor("got segment flange", timeout=5)
        proc_output.assertWaitFor("got segment tool0", timeout=5)
``` 

This might need some additional changes to confirm the given frames.

Also, concerning the multi robot checks, let me know if you need further adjustments as it is currently impossible to distinguish between the two robots for the "Robot initialized" test (all the other test cases check both robots as in those cases the 2 robots post different outputs)

PetoAdam:
okay, I will create a `humble` branch for the upstream repos too

kuka_sunrise_fri_driver/test/test_driver_activation.py
Svastits: Is it possible to ensure that the message arrived twice? that would be enough, if not, we should come back to this later

PetoAdam: I checked, it is possible to supply additional parameters, so by defining the process to check it is possible to distinguish between the 2 robots. The process names seem to be consistent both locally and on the ros-in-ci. However, I don't think that on the long run it would be an effective way of testing. eg:
```
proc_output.assertWaitFor("Robot initialized", process="robot_state_publisher-3", timeout=20)
proc_output.assertWaitFor("Robot initialized", process="robot_state_publisher-12", timeout=20)
```
So, I think checking for the same message twice is the way to go, even though it leaves something to be desired.

PetoAdam: So basically the assertWaitFor also checks previous logs, which means that calling it twice doesn't matter, the second call will also see the first "Robot initialized" message and pass

PetoAdam:
I created the humble branches for the upstream repos and also created a PR for adjusting the upstream.repos file on the humble branch in this repo to use the humble branch of the upstream repos. https://github.com/kroshu/kuka_drivers/pull/176

Still, updating the upstream repos for jazzy is needed before merging the PR.

Svastits (approved):

PR #178, opened 2024-08-06
Feature/fri external torque broadcast

krmihaly:
Added an external torque broadcaster for the sunrise FRI driver and a check for zero value in the `activateFrictionCompensation()` function.

### DIFF
--- a/kuka_sunrise_fri_driver/config/ros2_controller_config.yaml
+++ b/kuka_sunrise_fri_driver/config/ros2_controller_config.yaml
@@ -12,6 +12,8 @@ controller_manager:
       type: kuka_controllers/FRIStateBroadcaster
     event_broadcaster:
       type: kuka_controllers/EventBroadcaster
+    external_torque_broadcaster:
+      type: joint_state_broadcaster/JointStateBroadcaster
 
     # Configuration controllers
     joint_group_impedance_controller:
--- a/kuka_sunrise_fri_driver/launch/startup.launch.py
+++ b/kuka_sunrise_fri_driver/launch/startup.launch.py
@@ -39,6 +39,7 @@ def launch_setup(context, *args, **kwargs):
     jtc_config = LaunchConfiguration("jtc_config")
     jic_config = LaunchConfiguration("jic_config")
     ec_config = LaunchConfiguration("ec_config")
+    etb_config= LaunchConfiguration("etb_config")
     if ns.perform(context) == "":
         tf_prefix = ""
     else:
@@ -116,6 +117,7 @@ def launch_setup(context, *args, **kwargs):
             jtc_config,
             jic_config,
             ec_config,
+            etb_config,
             {
                 "hardware_components_initial_state": {
                     "unconfigured": [tf_prefix + robot_model.perform(context)]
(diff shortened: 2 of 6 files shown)


kuka_sunrise_fri_driver/src/hardware_interface.cpp
```diff
@@ -468,7 +468,9 @@ void KukaFRIHardwareInterface::activateFrictionCompensation(double * values) con
 {
   for (int i = 0; i < DOF; i++)
   {
-    values[i] -= (values[i] / fabs(values[i]) * 0.1);
+    if(values[i]!=0.0){
```
Svastits, line 471: please add else branch where 0 is also shifted with 0.1 (direction does not count for me)
this whole friction compensation might not make much sense, but before we verify that, edge cases should also be handled

kuka_sunrise_fri_driver/config/ros2_controller_config.yaml
```diff
@@ -12,6 +12,8 @@ controller_manager:
       type: kuka_controllers/FRIStateBroadcaster
     event_broadcaster:
       type: kuka_controllers/EventBroadcaster
+    external_torque_broadcaster:
```
Svastits, line 15: does this need a separate controller with the same type as the joint state broadcaster? would that not work to simply extend that config?

krmihaly: Not really, because the joint state broadcaster only broadcasts joint position velocity and effort. You can remap different hardware interface to the joint state (what i have done here with the external torque) but you cannot extend the state structure itself. 
Maybe I could create a completely new broadcaster very similar to the `jointstate` one with new message types and everything, but then it will not be compatible with other ROS nodes that might use the standard `jointstate` message.

Svastits: okay, thanks, you should definitely not create a new joint_state broadcaster, that would ruin compatibility, so your solution seems to be the right way
just to make sure that I understand corectly: you publish the joint state as previously on the joint_state topic and additionally publish a JointState message on the state topic, where the torque values are the external torques but all other values are the same?

krmihaly: Almost. The only data in it published to the "external_torque_broadcaster/joint_states" topic is a joint state message with the joint names and their external torques. The config yaml needs two extra lines to also broadcast the positions.

Svastits: great, sounds good :)

kuka_sunrise_fri_driver/launch/startup.launch.py
```diff
@@ -222,4 +225,11 @@ def generate_launch_description():
             + "/config/effort_controller_config.yaml",
         )
     )
+    launch_arguments.append(
```
Svastits, line 228: please add this new argument to the doc and also extend the started controllers list (+ information on the name of the topic and contents of the message)

Svastits (approved):

PR #184, opened 2024-09-12
Add Example for Depalletizing Using MoveIt Task Constructor

PetoAdam:
This Pull Request introduces a demonstration package for depalletizing tasks. The package leverages the MoveIt Task Constructor framework and is designed for use with KUKA robots.


PetoAdam:
Once the PR is merged, I will add the necessary documentation to the wiki.

examples/iiqka_moveit_task_constructor_example/kuka_moveit_task_constructor/launch/startup.launch.py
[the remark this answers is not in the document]
Svastits: or the standalone launch file + driver launch?

examples/iiqka_moveit_task_constructor_example/kuka_moveit_task_constructor/src/mtc_depalletizing_task_node.cpp
```diff
@@ -0,0 +1,314 @@
...
+{
+  rclcpp::init(argc, argv);
+
+  rclcpp::NodeOptions options;
+  options.automatically_declare_parameters_from_overrides(true);
+
+  auto mtc_depalletizing_task_node = std::make_shared<MTCDepalletizingTaskNode>(options);
+  rclcpp::executors::MultiThreadedExecutor executor;
+
+  auto spin_thread = std::make_unique<std::thread>(
```
Svastits, line 298: is there a reason for making this thread a unique ptr instead of simple var?

examples/iiqka_moveit_task_constructor_example/kuka_moveit_task_constructor/src/mtc_depalletizing_task_node.cpp
```diff
@@ -0,0 +1,314 @@
...
+  {
+    task.init();
+  }
+  catch (mtc::InitStageException & e)
+  {
+    RCLCPP_ERROR_STREAM(LOGGER, e);
+    return false;
+  }
+
+  if (!task.plan(1000))
```
Svastits, line 271: just out of curiosity, does this plan all steps at the same time? if yes, how is it possible that the robot does not collide with the upper pallets when going for the lower ones?

examples/iiqka_moveit_task_constructor_example/kuka_moveit_task_constructor/launch/startup_with_rviz_extended.launch.py
Svastits, line 1: I would not create an entire new launch file for just one different config file
You should either provide it as a launch arg or just delete one option, the user can modify the config file if he so desires.

In general 5 launch files are a bit too much for me, as they basically do the same with slight modifications, maybe even one could be enough with an argument whether to start rviz or not - I do not see the use case of the standalone launch files, as a driver is surely needed alongside, but you can convince me ;)

Svastits:
I do not think we can merge this to master currently, as mtc is not released and would therefore block our release, as the release process does not allow upstream repos. Maybe merge to rolling, but we should discuss this :)

examples/iiqka_moveit_task_constructor_example/kuka_moveit_task_constructor/src/mtc_depalletizing_task_node.cpp
PetoAdam: Some people like it, some does not. I sometimes use pointers for threading (mostly when they might throw an exception).
Advantages:
Automatic Cleanup: std::unique_ptr ensures the thread is cleaned up when it goes out of scope, preventing resource leaks.
Exception Safety: Ensures the thread’s destructor is called even if an exception is thrown, joining the thread if necessary.
Clear Ownership: Indicates single ownership, making the code easier to understand and maintain.

Disadvantages:
Overhead: May introduce unnecessary overhead in terms of performance and complexity.
Move Semantics: std::thread is already move-only, and wrapping it in a smart pointer can complicate the design.
Readability: Can reduce code readability and clarity by adding another layer of abstraction

Yes, I know it's a simple demo application and std::thread is pretty robust itself, I can change it to a simple var if you want. :)

examples/iiqka_moveit_task_constructor_example/kuka_moveit_task_constructor/launch/startup_with_rviz_extended.launch.py
PetoAdam: I knew it would be too many. I was just not sure, which ones to include.

I would personally keep the standalone launch file, because while the driver is waiting for setup and the robot manager's lifecycle is being configured and activated, the planner will try to create possible motion paths, which will all fail due to the driver not having been activated. The standalone version bypasses all this and makes it possible to independently run the motion planning and execution from the driver. Also, this makes it possible to run multiple different motion paths without having to restart the driver. (Also, it's independent from the driver configuration).

examples/iiqka_moveit_task_constructor_example/kuka_moveit_task_constructor/launch/startup.launch.py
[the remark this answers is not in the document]
PetoAdam: I would keep at least a separate launch file to keep the package its own unit. Or I'm not sure what you mean under reusing the example launch file. I would probably go with the latter. Standalone launch + driver launch. I just saw the example package's launch file had the driver launch inside it too, so I made some versions, which might make sense.

examples/iiqka_moveit_task_constructor_example/kuka_moveit_task_constructor/src/mtc_depalletizing_task_node.cpp
PetoAdam: It creates them one after the other. The console / rviz plugin shows the number of successful and failed solution for every step. It would be cool if it could plan all steps at the same time tho. :) (Too bad it would not know the exact start pose for some paths)

PetoAdam:
Yeah, I hope MTC will be available through apt soon(tm). We can discuss what to do tomorrow. :)

examples/iiqka_moveit_task_constructor_example/kuka_moveit_task_constructor/src/mtc_depalletizing_task_node.cpp
PetoAdam: Also, MTC is well integrated with the planning scene.

examples/iiqka_moveit_task_constructor_example/kuka_moveit_task_constructor/launch/startup_with_rviz_extended.launch.py
Svastits: okay, this perfectly makes sense, in the previous examples only the moveit server was started alongside the driver, the actual example/client was started after activation.
Then please keep the standalone file (rviz option might be an argument, or even 2 launch files are ok, if one uses the other)

examples/iiqka_moveit_task_constructor_example/kuka_moveit_task_constructor/src/mtc_depalletizing_task_node.cpp
Svastits: aaa, okay, this perfectly makes sense, (I just somehow missed the while loop)

examples/iiqka_moveit_task_constructor_example/kuka_moveit_task_constructor/launch/startup.launch.py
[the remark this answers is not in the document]
Svastits: by reusing I meant IncludeLaunchDescription, currently you use
```
    startup_launch = IncludeLaunchDescription(
        PythonLaunchDescriptionSource(
            [get_package_share_directory("kuka_iiqka_eac_driver"), "/launch/startup.launch.py"]
        ),
    )
```
but as I saw, the example launch file with the moveit server could also be used instead of this

BUT based on the launch file discussion, you should go with the standalone launch file, where no include is possible

examples/iiqka_moveit_task_constructor_example/kuka_moveit_task_constructor/src/mtc_depalletizing_task_node.cpp
PetoAdam: the custom deleter is a better approach, as it can gracefully stop a thread, while in the case of a smart pointer, if the pointer runs out of scope, the thread's destructor is called, which terminates the thread abruptly. That's why I manually joined the thread in the example. Having a custom deleter would be more ideal, I just did not think this simple example would need it.

examples/iiqka_moveit_task_constructor_example/kuka_moveit_task_constructor/launch/startup.launch.py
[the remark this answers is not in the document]
PetoAdam: okay, I'll then keep the standalone ones. :)

examples/iiqka_moveit_task_constructor_example/kuka_moveit_task_constructor/src/mtc_depalletizing_task_node.cpp
Svastits: But compared to the simple std::thread, what is the difference without the custom deleter? the simple var would also run out of scope and the destructor would be called. So I do not see why the unique ptr is better in this case, as without the deleter it will terminate the same way (also in case of an exception, where join() would not be called automatically)

(I do not necessarily want to change to simple var, just want to understand the effects of both and without a custom deleter I think your first two advantages are not valid, as the behaviours of both cases are identical)

PetoAdam: I read a bit about how exactly it works. Yes, you are right, the only reason one might want to use a smart pointer for storing a thread is for ownership purposes. It does not really apply for our case, so changing it back to a simple var makes a lot of sense.

Svastits:
Closing this, migrated to examples repo: https://github.com/kroshu/examples/pull/11

PR #185, opened 2024-10-02
ROS2 control hwif and launch config update

PetoAdam:


kuka_iiqka_eac_driver/launch/startup.launch.py
```diff
@@ -145,29 +145,37 @@ def launch_setup(context, *args, **kwargs):
...
         return Node(package="controller_manager", executable="spawner", arguments=arg_list)
 
-    controller_names = [
-        "joint_state_broadcaster",
-        "joint_trajectory_controller",
-        "joint_group_impedance_controller",
-        "effort_controller",
-        "control_mode_handler",
-        "event_broadcaster",
+    controllers = {
```
Svastits, line 166: I believe you, that providing the config files to the spawner too was necessary, but do you know why?

not so long ago, the structure was similiar, but then each config file was only provided to the appropriate spawner, not to the controller_manager. 
But then there was a change in humble, where the default values were loaded by the controller manager before the spawner could provide the parameters, which resulted in errors in all controllers.

Was this change reverted? could you check what happens, if the controller_manager does not get the config files?

PetoAdam: I'm not entirely sure. Without providing the config files to the spawner, I got the error (both locally and on the ci) that the given controllers (defined in the ros2_controllers_config.yaml) were found, but not their specific parameters. So I had to provide them to the spawners. I also think this might be a cleaner way to provide configs for the controllers.

Yes, I checked. This way you can remove the controller specific config files from the control_node. But, the main ros2_controller_config still has to be provided to the control_node, which makes sense. I guess, I can remove the controller specific config parameters from the control node, but please confirm this. :)

PetoAdam: Okay, I removed the controller config files from the control_node as they are not needed.

Svastits (approved):

PR #204, opened 2024-12-16
Fix event_broadcaster topic name

Svastits:


Svastits:
@mergifyio backport humble

Issue #211, opened 2025-02-24
Feature request: custom controller name for KSS drivers


GiuseppeMonetti95:
In the current implementation of the ROS2 drivers, it is not possible to have a controller for KSS drivers that has a name different from `joint_trajectory_controller`.

This happens because the static variable `kuka_drivers_core::JOINT_TRAJECTORY_CONTROLLER` is used in the KSS robot manager, in particular:
- In [on_activate()](https://github.com/kroshu/kuka_drivers/blob/9c3f1e945f7a00a8000dc79dc56c7893c690ec03/kuka_kss_rsi_driver/src/robot_manager_node.cpp#L102)
- In [on_deactivate()](https://github.com/kroshu/kuka_drivers/blob/9c3f1e945f7a00a8000dc79dc56c7893c690ec03/kuka_kss_rsi_driver/src/robot_manager_node.cpp#L131)

On the other side, both Sunrise and iiQKA drivers have this feature.

PR #212, opened 2025-02-24
Feature/kss controller name parameter

GiuseppeMonetti95:
With this PR, the feature requested in issue #211 is implemented.

Simply, a new ROS2 parameter is added to the robot manager of KSS drivers.

To provide this parameter, a config file `driver_config.yaml` has been added (similarly to what happens in both iiQKA and Sunrise drivers).

Moreover, the new parameter and config file have been documented in the wiki.


kuka_kss_rsi_driver/include/kuka_kss_rsi_driver/robot_manager_node.hpp
```diff
@@ -56,6 +56,7 @@ class RobotManagerNode : public kuka_drivers_core::ROS2BaseLCNode
   rclcpp::CallbackGroup::SharedPtr cbg_;
 
   std::string robot_model_;
+  std::string controller_name_;
```
Svastits, line 59: please rename this to show that it is the name of the joint position controller, e.g. `joint_pos_controller_name_` or `position_controller_name_` would both do

Svastits:
Approved after variable rename and pre-commit format fix (this can be done automatically with `pre-commit run -a` after installig it)

kuka_kss_rsi_driver/include/kuka_kss_rsi_driver/robot_manager_node.hpp
GiuseppeMonetti95: Done!

PR #225, opened 2025-04-11
IO handling for RSI only KSS driver

altex111:
This pull request adds IO control solution for KSS driver
 - This is only for RSI only solution
 - Important to note that this pull request adds only compile time GPIO configuration. Configuration file IO config will be in a later PR
 - The GPIOs can be configured in kuka_external_control_sdk/kss/message_builder.h
 
 CI should fail since it pulls packages from master

### DIFF
--- a/doc/wiki/3_Sunrise_FRI.md
+++ b/doc/wiki/3_Sunrise_FRI.md
@@ -8,7 +8,7 @@
 
 #### Controller side
 
-- Upload the robot application under `robot_application/src` to the controller using Sunrise Workbench
+- Upload the robot application under `robot_application/src` to the controller using Sunrise Workbench. The driver only supports Sunrise 1.X.
 
 ### Configuration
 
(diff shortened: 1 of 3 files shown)


kuka_kss_rsi_driver/src/hardware_interface_rsi_only.cpp
```diff
@@ -168,7 +210,32 @@ void KukaRSIHardwareInterface::Read(const int64_t request_timeout)
   {
     const auto & req_message = robot_ptr_->GetLastMotionState();
     const auto & positions = req_message.GetMeasuredPositions();
+    const auto & gpio_values = req_message.GetGPIOValues();
+
     std::copy(positions.cbegin(), positions.cend(), hw_states_.begin());
+    // Save IO states
+    for (size_t i = 0; i < hw_gpio_states_.size(); i++)
+    {
+      switch (gpio_values.at(i)->GetGPIOConfig()->GetValueType())
```
pasztork, line 219: Maybe create a `GetValueAsDouble()` method. That way there will be no need for the switch statement.

pasztork (approved):

PR #237, opened 2025-07-15
Fix/fix note for release

altex111:
Fixing a deprecated include to clean the code:
image

- [x] Code builds


altex111:
On wsl this fixes the note, but on a linux computer this makes it fail to build.

PR #307, opened 2026-02-17
Cleanup package.xml dependencies

Svastits:
colcon also enforces exec_dependencies to be available when building a package, which disables deploying only a few packages to machine with limited resources.
Therefore dependencies that are not strictly mandatory, have been removed in this PR


kuka_rsi_driver/package.xml
```diff
@@ -20,17 +19,14 @@
   <depend>kuka_drivers_core</depend>
   <depend>pluginlib</depend>
   <depend>std_msgs</depend>
-
   <depend>tinyxml_vendor</depend>
 
   <exec_depend>joint_trajectory_controller</exec_depend>
   <exec_depend>joint_state_broadcaster</exec_depend>
-  <exec_depend>gpio_controllers</exec_depend>
```
pasztork: Are you sure we don't need this?

PR #310, opened 2026-02-19
Change model verification to only consider payload, reach and type

pasztork:


gavanderhoorn:
To keep things connected: https://github.com/kroshu/kuka_drivers/pull/309#issuecomment-3926944264

Svastits (approved):

PR #335, opened 2026-03-25
Call controller `update` with fixed period to avoid interpolation jitter

Svastits:


pasztork (approved):

destogl:
Dear maintainers, just FYI:

Recently we have added `slave` hardware interface `async` type and also are enabling the upstream `ros2_control_node` to not sleep, but to get synced to the robot controller. This solves the jitter when only one robot is controlled. For multi-robot we are on it, but it will take time.

I hope that [this PR](https://github.com/ros-controls/ros2_control/pull/3213) will make this PR and other your efforts for it obsolete and you will be able to use the upstream node. If you could test that approach too, your input would be very valuable for us :)

Thank you!

Svastits:
@destogl  thanks for the hint :) do you already know to which versions will this be backported?

PR #350, opened 2026-07-20
Stabilize driver activation tests

Svastits:
## Summary by CodeRabbit

- **New Features**
  - Added a `lifecycle_manager` executable that drives a target ROS 2 node through configure and activate transitions using configurable startup delays.
  - Transition progress and failures are surfaced via logs.

- **Tests**
  - Updated driver activation tests to launch the new `lifecycle_manager` instead of invoking lifecycle commands/timers.
  - Switched test launch imports/dependencies to `launch_ros` (removing `ros2lifecycle`).

### DIFF
--- a/kuka_drivers_core/src/lifecycle_manager.cpp
+++ b/kuka_drivers_core/src/lifecycle_manager.cpp
@@ -46,8 +46,8 @@ public:
     configure_delay_ = this->get_parameter("configure_delay").as_double();
     activate_delay_ = this->get_parameter("activate_delay").as_double();
 
-    change_state_client_ = this->create_client<lifecycle_msgs::srv::ChangeState>(
-      managed_node + "/change_state");
+    change_state_client_ =
+      this->create_client<lifecycle_msgs::srv::ChangeState>(managed_node + "/change_state");
 
     worker_thread_ = std::thread([this]() { this->run(); });
   }
--- a/kuka_drivers_core/src/lifecycle_manager.cpp
+++ b/kuka_drivers_core/src/lifecycle_manager.cpp
@@ -49,6 +49,8 @@ public:
     change_state_client_ =
       this->create_client<lifecycle_msgs::srv::ChangeState>(managed_node + "/change_state");
 
+    // Run delayed transitions in a worker thread so the executor can keep spinning
+    // and respond quickly to shutdown while timing logic runs independently.
     worker_thread_ = std::thread([this]() { this->run(); });
   }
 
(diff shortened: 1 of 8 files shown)


kuka_drivers_core/src/lifecycle_manager.cpp
```diff
@@ -0,0 +1,126 @@
...
+    if (remaining_delay > 0.0)
+    {
+      std::this_thread::sleep_for(std::chrono::duration<double>(remaining_delay));
+    }
+
+    RCLCPP_INFO(this->get_logger(), "Sending 'activate' transition");
+    if (!changeState(lifecycle_msgs::msg::Transition::TRANSITION_ACTIVATE))
+    {
+      RCLCPP_ERROR(this->get_logger(), "'activate' transition failed");
+    }
```
pasztork, line 146: Maybe also add a warning if the configured delays couldn't be held.

kuka_drivers_core/src/lifecycle_manager.cpp
```diff
@@ -0,0 +1,126 @@
...
+    this->declare_parameter<double>("activate_delay", 25.0);
+
+    const auto managed_node = this->get_parameter("managed_node").as_string();
+    configure_delay_ = this->get_parameter("configure_delay").as_double();
+    activate_delay_ = this->get_parameter("activate_delay").as_double();
+
+    change_state_client_ =
+      this->create_client<lifecycle_msgs::srv::ChangeState>(managed_node + "/change_state");
+
+    worker_thread_ = std::thread([this]() { this->run(); });
```
pasztork, line 54: Why use a different thread?

pasztork (changes requested):

kuka_drivers_core/src/lifecycle_manager.cpp
Svastits: what do you mean by could not be held?

kuka_drivers_core/src/lifecycle_manager.cpp
Svastits: // Run delayed transitions in a worker thread so the executor can keep spinning
    // and respond quickly to shutdown while timing logic runs independently.

kuka_drivers_core/src/lifecycle_manager.cpp
pasztork: When `remaining_delay < 0`. It might be a sign that something got stuck or some node just didn't show up. This can lead to hanging processes.

Svastits: `remaining_delay` is currently only calculated from the two configured values, so in normal setup will never be negative

pasztork (approved):

